PR #57, opened 2015-02-01
Give PEAR it's own ensure.

smoriahNYT:
php::pear currently defaults to ensure "$php::ensure". This causes some
trouble once versions are specified as PHP could be on 5.6 and PEAR at 1.9.


fpletz:
Good catch! Thanks! :beers:

PR #64, opened 2015-03-11
Re-adding support for PHP < 5.4 (eg, Ubuntu 12) without breaking 5.4+

alivesay:
We're still running some Ubuntu 12.04 systems and need support for both PHP 5.3 and 5.4+ from puppet-php.  These changes seems to do the trick.

### DIFF
--- a/lib/facter/php_version.rb
+++ b/lib/facter/php_version.rb
@@ -1,0 +1,11 @@
+Facter.add(:php_version) do
+  confine :kernel => :linux
+  setcode do
+    version = Facter::Util::Resolution.exec("/usr/bin/env php -r 'echo PHP_VERSION;' 2>/dev/null")
+    if version
+      version.match(/(\d+\.\d+\.\d+)/).to_s
+    else
+      nil
+    end
+  end
+end
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -134,16 +134,32 @@ define php::extension(
   # Ubuntu/Debian systems use the mods-available folder. We need to enable
   # settings files ourselves with php5enmod command.
   if $::osfamily == 'Debian' {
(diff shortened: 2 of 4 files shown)


fpletz:
@alivesay Well, that sucks. :crying_cat_face:

Did you see the comments in #44? We decided against supporting PHP < 5.4 because it adds unnecessary complexity and PHP 5.3 is not supported by upstream anymore. Additionally, due to the custom fact `php_version`, at least two Puppet runs are necessary for applying the configuration. This is not a proper solution in our opinion.

I'm very sorry that we have to reject this in this state. Do you maybe have an idea how to achieve this in a more elegant fashion?

alivesay:
@fpletz Ah I did miss #44.  Hmm..

What if I dropped the custom `php_version` fact and just looked for the presence of `$php::params::config_root/mods-available`?

Thanks.

globin:
I'm actually inclined to close this, due to the fact that I don't want to support the use of unsafe, outdated versions. I'd like to ask you to either try and use a newer version which shouldn't be a major problem or to keep a fork temporarily.

If we really get more requests to support 5.3 again I'll reconsider my current opinion :panda_face: 

But definitely thanks for your work :beers: 

(@fpletz, @mmoll ?)

mmoll:
Personally I would like to keep the fact gone but 5.3 support would be nice, as long as it doesn't really add huge complexity.

fpletz:
A solution without the fact and without hackish checks that infer the PHP version by the existence of distribution-specific files would be preferable but that this probably won't be possible.

Thanks for your work and your comments, but I'm afraid we won't merge this. I'm closing this PR but feel free to reopen if you have an alternative solution. :cookie:

PR #70, opened 2015-03-18
get closer to Puppet4 readyness

mmoll:
Comments please :newspaper:, there are two tests failing, maybe @globin has an idea about the inifile paramter, and I still have to think over the spaceship operator, as long as nobody else has a quick solution for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: ruby
 rvm:
 - 2.1.5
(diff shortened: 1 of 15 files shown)


globin:
Is it possible to turn of the empty string lint for that line? We actually want that there because of php.ini weirdness having no section/an empty section in a few places.

And we always want to make sure we always restart fpm after installing/changing extensions even if we don't define them. Hence the empty spaceship is fully intended.

Are those not allowed in Puppet4 or is that just fancy "hack-freeness"?

mmoll:
I don't think it's possible to turn off the check for just one line. See http://www.camptocamp.com/actualite/getting-code-ready-puppet-4/ for details on the checks. I would be OK to turn off the spaceship check if there's no good solution, especially as I used it in the past, too. ;)

globin:
http://puppet-lint.com/controlcomments/

mmoll:
Ah, nice, didn't know that. :bow:
However, would ```undef``` instead of ```''``` also do the trick for inifile?

globin:
If it's allowed here: https://github.com/puppetlabs/puppetlabs-inifile/blob/master/lib/puppet/util/ini_file.rb#L12

I don't know, someone ( :panda_face: ) will have to try,

mmoll:
doesn't work with ```undef``` :crying_cat_face:

mmoll:
All :green_heart: now, for your kind consideration.
BTW, we really need more spec tests, I don't think this is really future parser ready...

globin:
Go for it :wink:

PR #75, opened 2015-04-21
PHP 5.4 on Ubuntu Precise and improved Pecl/zend_extension handling

adepretis:


fpletz:
Thanks for your pull request!

I've cherry-picked https://github.com/25th-floor/puppet-php/commit/59f8f02cd1dc4d2501cbcefb4028261f2ad6964b into master. :+1:

However, what's the rationale for adding the full path to the extensions? Shouldn't php be able to load those modules anyway? Additionally, adding a custom fact would require a second puppet run. We should try to avoid this.

adepretis:
Thx for your feedback @fpletz 

Only PHP 5.5+ supports `zend_extension` without a full path to the extension directory (see http://git.php.net/?p=php-src.git;a=commitdiff;h=0b8b6a727ddd31ff14e4af919c77a3f1b5e2b3f0). To fully use this great Puppet module with `ondrej/php5-oldstable` using an absolute path is therefore required.

f2e7bfb also contains an update to the modules pecl integration. Because e.g. for Zend OPcache the pecl package name is `zendopcache` but the resulting DSO is called `opcache.so`. I don't know if this is the only example, but again, for PHP 5.4 to be fully supported by this Puppet module, it's an requirement.

Regarding the "second puppet run" for custom facts I don't know exactly what you mean. Many modules (including some of the deps of `mayflower-php`) use custom facts, and I see no problems with that.

fpletz:
Thanks for your explanations. Agreed! :+1:

@globin What do you think about the naming of the parameter `pecl_name`?

I think this would break backwards-compatibility because the package name in pecl was previously the title of an `php::extension`. With these changes, that title would be the name of the shared object instead. Shouldn't we inverse that logic and use an `so_name` parameter instead?

@adepretis I don't have problem with custom facts, but in this case that custom fact could cause failed first puppet run and would require at least a second puppet run: Facts are generated before the puppet agent applies the catalog. At this stage, we cannot guarantee that php-cli is installed and therefore the execution of `/usr/bin/env php` will fail. The default `20100525` is sensible and used by all PHP versions we support, but this might bite us in the future.

Therefore, I would prefer configuring `extension_version` manually in hiera instead of adding a broken autodetection mechanism. @globin What's your take on this?

adepretis:
Funny, I went for the `so_name` approach first, and refactored it to pecl_name afterwards ... for some reason it made more sense, but don't ask me why/what ;-)

Regarding `extension_version` / custom fact: ok, makes sense, totally forget this case. I'll try to refactor it into a hiera setting on the weekend.

adepretis:
Ok, I updated the PR to use so_name instead of pecl_name and php_api_version instead of the custom fact. Wether the full path is used for zend_extension depends solely on php_api_version being set. Any thoughts on this?

manifests/extension.pp
```diff
@@ -22,10 +22,17 @@
...
+#
+# [*php_api_version*]
+#   This parameter is used to build the full path to the extension
+#   directory for zend_extension in PHP < 5.5 (e.g. 20100525)
+#
 # [*header_packages*]
 #   System packages dependencies to install for extensions (e.g. for
 #   memcached libmemcached-dev on debian)
-#
+#
```
globin: Additional whitespace

globin:
Except for the nit, LGTM.

@fpletz ack?

manifests/extension.pp
```diff
@@ -42,6 +49,8 @@
   $provider          = undef,
   $source            = undef,
   $pecl_source       = undef,
+  $so_name           = $name,
+  $php_api_version   = undef,
```
fpletz: We should probably move the `$php_api_version` parameter into the main class, because that setting would apply to all php extensions as we don't support multiple versions of PHP on a system. In case the api version needs to be overridden, every extension hash in hiera would need to specify `php_api_version` which is a bit redundant.

So my suggestion would be to move `$php_api_version` to the `php` class as `$api_version` and use `${::php::api_version}` where `$module_path` is built.

adepretis:
@fpletz I guess I made a mistake by doing a merge commit (1ff2447) ... squashing and rebasing to the current master seems to be affected by that

fpletz:
@adepretis No problem, I've gone ahead and rebased + squashed your changes. Additionally, I fixed the commit description because it wasn't up to date anymore. See https://github.com/mayflower/puppet-php/commit/85919344312271047f4afe1c9963f8d3eea4f253 for reference.

Thanks for your contributions! :beers:

PR #77, opened 2015-04-28
$ensure handling for module and extensions

codec:
I was trying to pin to a specific PHP version and encountered some problems with the module.

This is the node data I was using (i.e. `common.yaml` in wasted):
```yaml
php::ensure: 5.5.9+dfsg-1ubuntu4.9
php::fpm::ensure: 5.5.9+dfsg-1ubuntu4.9
php::cli::ensure: 5.5.9+dfsg-1ubuntu4.9
php::manage_repo: false

component::php::manage_repo: true
component::php::repo::repo:
  jenkins_php5.5.5.9:
    key: F5067765
    key_source: https://jenkins.example.com/php/PUBKEY.gpg
    location: http://jenkins.example.com/php/
    repos: main
    include_src: false
```
This is supposed to pin to 5.5.9, but some modules have different versions than the PHP core (see commit message) resulting in a broken install (package/version not available).

I've worked around this by creating all extensions outside of the module and then passing all the exts into the component, like this:
```puppet
class component::php (
  $extensions = {}
) {
  contain ::php

  create_resources('::php::extension', $extensions, { ensure => present })
 }
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -107,7 +107,6 @@ class php (
 
   $real_extensions = hiera_hash('php::extensions', $extensions)
   create_resources('::php::extension', $real_extensions, {
-    ensure  => $ensure,
     require => Class['::php::cli'],
     before  => Anchor['php::end']
   })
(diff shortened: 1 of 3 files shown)


fpletz:
:+1: :beers:

PR #80, opened 2015-05-04
Bugfix: changed - to _ to match default value yum::repo::remi_php56

netors:
See:
https://github.com/example42/puppet-yum/blob/master/manifests/repo/remi_php56.pp#L5


fpletz:
Thanks! :beers:

PR #82, opened 2015-05-06
Ensure extension setting prefix

### LINKED ISSUE #81: Ensure extension settings prefix
Currently extension settings have to specified like this:

```yaml
php::extensions:
  apc:
    package_prefix: php-
    settings:
      apc.stat: 1
      apc.stat_ctime: 1
```

A bit more intuitive and de-duplicating would be this:

```yaml
php::extensions:
  apc:
    package_prefix: php-
    settings:
      stat: 1
      stat_ctime: 1
```

To provide the still-required extension prefix, the [`prefix()`](https://github.com/puppetlabs/puppetlabs-stdlib#prefix) function from `puppetlabs-stdlib` could be used but in that case existing configuration would lead to something like `apc.apc.stat` and break. Thus a separate function could be necessary to only add the prefix if not present.

Any opinions if this is worth adding?

mbrodala:
By default extension setting keys are now prefixed with `<name>.`
If the prefix is already present, nothing is changed.

This allows for simpler extension configuration by simply passing
the setting name without prefix.

Fixes #81

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -33,6 +33,10 @@
 # [*settings*]
 #   Nested hash of global config parameters for php.ini
 #
+# [*settings_prefix*]
+#  Boolean parameter, whether to prefix all setting keys with
+#  the extension name. Defaults to true.
+#
 define php::extension(
   $ensure            = 'installed',
   $provider          = undef,
@@ -43,6 +47,7 @@ define php::extension(
   $compiler_packages = $::php::params::compiler_packages,
   $zend              = false,
   $settings          = {},
+  $settings_prefix   = true,
 ) {
 
   if $caller_module_name != $module_name {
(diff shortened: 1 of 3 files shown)


mbrodala:
I'll fix the test breakages.

manifests/extension.pp
```diff
@@ -43,6 +47,7 @@
   $compiler_packages = $::php::params::compiler_packages,
   $zend              = false,
   $settings          = {},
+  $settings_prefix   = true,
```
globin: I think we should default to false, at least until we release a new major version, due to the fact that not all extensions use their name as prefix (APCu uses `apc.`)

manifests/extension.pp
```diff
@@ -110,23 +115,36 @@
 
   $lowercase_title = downcase($title)
 
+  # Ensure "<extension>." prefix is present in setting keys if requested
+  if $settings_prefix {
+    $full_settings = ensure_prefix($settings, "${name}.")
```
globin: This should probably be configurable optionally for cases like the one mentioned above. Using the `$lowercase_title` is probably desired to.

manifests/extension.pp
mbrodala: OK, done.

manifests/extension.pp
mbrodala: I made `settings_prefix` boolean/string now.

fpletz:
Awesome work! Thanks! :beers:

PR #85, opened 2015-05-12
Remove pool's default listen_allowed_clients and add conditional in template

arielsalvo:
The only way to open the service to any IP is to not include the `listen.allowed_clients` option in the pool. (Since I want to manage this through AWS security groups, it is still safe)
Can't set an empty value in YAML, apparently.

**WARNING**: This means a change in behaviour!!!
Those that set `listen` to something other than 127.0.0.1 and trust the default to allow only requests sourced from 127.0.0.1, will find this change allows requests from any source.

### DIFF
--- a/manifests/fpm/pool.pp
+++ b/manifests/fpm/pool.pp
@@ -93,7 +93,7 @@ define php::fpm::pool (
   $listen = '127.0.0.1:9000',
   # Puppet does not allow dots in variable names
   $listen_backlog = '-1',
-  $listen_allowed_clients = '127.0.0.1',
+  $listen_allowed_clients = undef,
   $listen_owner = undef,
   $listen_group = undef,
   $listen_mode = undef,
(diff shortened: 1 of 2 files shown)


fpletz:
Thanks! We should definitely make `listen_allowed_clients` optional. :+1:

Actually, I think that it doesn't make much sense to modify the listening IP without changing `listen_allowed_clients`. For instance, why would you listen to `0.0.0.0` without changing `listen_allowed_clients`?

As we default to listen to `127.0.0.1 `, I'm okay with not setting `listen_allowed_clients` by default.

We should mention this in the changelog, though, but in my opinion we don't need a new major version for this change. @globin?

arielsalvo:
I don't see why anyone would either.... but there are crazies everywhere :)
Since you CAN do it and the result would be quite insecure, I thought I'd bring it up.

fpletz:
Totally forgot about this... let's merge it. Thanks again! :beers:

PR #96, opened 2015-06-25
Allow sending errors to syslog

### LINKED ISSUE #95: Can't send errors to syslog in php-fpm.conf
It's not possible to set `error_log = syslog`

If this ability is added, it should also be possible to configure the syslog facility or identity (`syslog.facility` and `syslog.ident`)

3flex:
Allows for any value to be set for error_log, and adds configuration for syslog.ident and syslog.facility.

Fixes #95

### DIFF
--- a/manifests/fpm/config.pp
+++ b/manifests/fpm/config.pp
@@ -24,6 +24,10 @@
 #   Whether to purge pool config files not created
 #   by this module
 #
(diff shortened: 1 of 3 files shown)


fpletz:
Perfect work! Thanks! :beers:

PR #106, opened 2015-07-24
Add check if value is empty in the pool.conf.erb template

ironpinguin:
The php fpm failed to start if the configuration for a pool contains something like 
`env[FOOBAR] = `

So with this fix the defect configuration will be prevent.


fpletz:
Does not compute. If you're happy and you know it, syntax error. :boom:

fpletz:
:+1:

PR #115, opened 2015-09-07
Embedded SAPI support (php.ini)

glorpen:
Support for embedded sapi - configuring specific for sapi `php.ini` as it is only thing I think is needed in this case.
Checked on RedHat and Debian. Fingers crossed for Suse & FreeBSD.


fpletz:
Sorry it took to so long to review this. Looks good. :+1:

PR #116, opened 2015-09-09
Include repositories for osfamily Debian, not just wheezy

codec:
Repos work fine for the entire family of Debian-based systems.


globin:
Then debian has to be fixed for releases != wheezy, currently all supported Ubuntu versions work and Debian wheezy work.

codec:
Oops. Updated to explicitly state manage_repos behavior for Debian-ish systems..

fpletz:
This bug was introduced by my commit 2ad20dd86d2a019c7bb464511f9b81d99a767dcf. :sheep: 

Looks good, thanks! :smoking:

PR #117, opened 2015-09-11
Create upstart service override before installing php-fpm

fpletz:
@globin


manifests/fpm.pp
```diff
@@ -49,4 +49,14 @@
 
   $real_pools = hiera_hash('php::fpm::pools',  $pools)
   create_resources(::php::fpm::pool, $real_pools)
+
+  # Create an override to use a reload signal as trusty and utopic's
+  # upstart version supports this:q
```
mmoll: :q!

fpletz: :panda_face:

PR #119, opened 2015-09-23
PHP custom facts

codec:
I guess only the `phpversion` fact is actually useful, but it's hacky.

### DIFF
--- a/lib/facter/phpversion.rb
+++ b/lib/facter/phpversion.rb
@@ -1,0 +1,7 @@
+Facter.add(:phpversion) do
+  setcode do
+    Facter::Util::Resolution.exec('php -v').
+      split("\n").first.split(' ').
+      select { |x| x =~ /^(?:(\d+)\.)(?:(\d+)\.)?(\*|\d+)/ }.first
+  end
+end


fpletz:
We've rejected and even removed custom facts in the past because they need a second Puppet run to be available in the scope. Thus, I'm not sure if this is really necessary. What is the use case?

lib/facter/fpm_conf_d.rb
```diff
@@ -0,0 +1,11 @@
+Facter.add(:fpm_conf_d) do
+  confine :osfamily => 'Debian'
+
+  setcode do
+    Dir.glob('/etc/php5/fpm/conf.d/*.ini').each do |dot_config|
```
stupied4ever: it will not work on RetHat distros as set on params.pp, the fpm configuration leaves at:
/etc/php-fpm.d/

lib/facter/fpm_pools.rb
```diff
@@ -0,0 +1,9 @@
+Facter.add(:fpm_pools) do
+  confine :osfamily => 'Debian'
+
+  setcode do
+    Dir.glob('/etc/php5/fpm/pool.d/*.conf').each do |pool_config|
```
stupied4ever: it will not work on RetHat distros as set on params.pp, the fpm configuration leaves at:
/etc/php-fpm.d/

stupied4ever: I know this is only on debian, but if merge that, i think the should be this feature on all distros.

lib/facter/fpm_conf_d.rb
codec: Thats why it's confined to Debian.

codec:
Having the `phpversion` fact easily allows checks against PuppetDB (are we running the same version on all servers, etc).

codec:
How about only cherry-picking 094fc9d for now? There's probably no real use for the other facts. For `fpm::pool` it's even sufficient to query PuppetDB for catalogs instead.

globin:
Especially phpversion has caused a lot of problems for people not directly involved with this module, trying to use it to make sure to install different things based on the php version e.g. apc for php <= 5.4 apcu for php > 5.4. This will fail on the first puppet run.

I'd like some more input from the other @mayflower/puppeteers?

codec:
@globin Clearly state in the README that the `phpversion` fact is not to be used for something like this? (If you do, you obviously don't understand how it works)

mmoll:
I can see the win to have this in a PuppetDB or Foreman inventory, so :+1:, but this would need more testing on more platforms, e.g. I'm unsure if that works as-is on FreeBSD as `php` is in /usr/local/bin which might not be in `$PATH` per default.

fpletz:
@codec Yes, please rebase to add just the `phpversion` fact and we'll merge it for now.

codec:
@fpletz :heavy_check_mark: : 

I agree with @mmoll. No clue how we should go about this, though.
`Facter::Util::Resolution.exec` already has a bit of error handling, but it does so quietly.
See https://downloads.puppetlabs.com/facter/apidocs/classes/Facter/Util/Resolution.html#M000070

mmoll:
well, I'd say just go with it and wait if @fraenki cries ;)

PR #122, opened 2015-09-28
Allow multiple releases to be deployed.

MG2R:
Also generalized the inclusion of dotdeb-wheezy.
Backwards compatibility has remained.

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -6,7 +6,8 @@
 #   Location of the apt repository
 #
 # [*release*]
-#   Release of the apt repository
+#   Releases of the apt repository. Can be a string with the release name, or an
+#   array of strings with multiple releases.
 #
 # [*repos*]
 #   Apt repository names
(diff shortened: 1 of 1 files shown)


manifests/repo/debian.pp
```diff
@@ -42,24 +43,30 @@
...
     location    => $location,
-    release     => $release,
     repos       => $repos,
     include_src => $include_src,
-    require     => Apt::Key['php::repo::debian'],
   }
 
   if ($dotdeb) {
-    # wheezy-php55 requires both repositories to work correctly
+    # Dotdeb requires the general 'wheezy' release as well.
```
fpletz: Are you sure? Even on jessie? The instructions don't reflect this.

fpletz:
Hmm, I don't quite understand what you mean with 'multiple releases'. `php::repo::debian` is still a class so there is just one apt source that's being deployed.

MG2R:
Please ignore this pull request. I don't know what I was thinking when I wrote this, but I just rolled this back on my own environments.

fpletz:
Thanks for the follow-up.

PR #123, opened 2015-09-28
Include fpm pid file configuration

stupied4ever:
Include a option to configure the fpm pid file. As in some distros (and
pgp versions) the pid file on the init script changes, we need to
provide a way to the user override the fpm pid file.

For example, at RedHat (AmazonLinux) the php56 package leves at:
`
/var/run/php-fpm/php-fpm-5.6.pid
`

With this change, the Amazon Linux user with the php56 package should
configure on hiera the key:

```yaml
php::fpm::config::pid_file: /var/run/php-fpm/php-fpm-5.6.pid
```


fpletz:
Thanks! :+1:

PR #124, opened 2015-10-07
Allow removal of config values

mpdude:
With this change, config values can be removed entirely.

For example, this will remove the `date.timezone` configuration entry:

```puppet
class { '::php':
  settings => {
    'Date/date.timezone' => undef
  }
}
```

### DIFF
--- a/manifests/config/setting.pp
+++ b/manifests/config/setting.pp
@@ -50,6 +50,6 @@ define php::config::setting(
     value   => $value,
     path    => $file,
     section => $section,
-    setting => $setting
+    setting => $setting,
   }
 }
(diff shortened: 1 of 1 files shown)


globin:
Why did you remove the section?

mpdude:
Oops, that happens when you think a change is so small you can make it in the GitHub web editor. Will fix that.

globin:
Oh and probably the build will fail because ensure is not the first parameter %)

mpdude:
Updated.

mpdude:
Thanks, I'll give you a :star: for that :smiley:

globin:
Thanks for the PR :beers:

PR #125, opened 2015-10-08
Make sure auto_discover is set *before* installing packages

mpdude:
I may be totally wrong here, but shouldn't this be set up before the packages are installed?

For me, package installation fails (on a completely fresh box) with the hint to use "pear channel-discover pear...." to initialize or pear config-set auto_discover 1".

### DIFF
--- a/lib/puppet/provider/package/pear.rb
+++ b/lib/puppet/provider/package/pear.rb
@@ -85,7 +85,7 @@ Puppet::Type.type(:package).provide :pear, :parent => Puppet::Provider::Package 
   end
 
   def install(useversion = true)
-    command = ["upgrade"]
+    command = ["-D", "auto_discover=1", "upgrade"]
     if @resource[:install_options]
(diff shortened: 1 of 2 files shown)


fpletz:
We don't really use pear, so we didn't stumble on this. Looks good. :+1:

mpdude:
"Legacy", you know... What is your policy for tagging new releases?

fpletz:
Right now our focus is to add some more possibly breaking changes for 4.0.0, some of which are already in the current master branch. As this is actually a bug fix it would make sense to release a new point release which is probably what you're asking for? :panda_face:

mpdude:
Oh wait – I'll have to check whether puppet-librarian is working with the current master or a stable release :-)

mpdude:
Just checked that. As I also need the #124 feature (which is on `master`) I'll stick with that and wait for the official 4.0.0.

Thanks!

PR #128, opened 2015-10-12
Enable extension when package already installed

mpdude:
The `php5query` command needs a SAPI. We use `cli` here and hope that all extensions are managed for all SAPIs consistently.


mpdude:
*ping* @fpletz / @globin

fpletz:
Seems reasonable as we're currently enabling extensions for all SAPIs and `cli` is always installed by default. Thanks! :+1:

Issue #129, opened 2015-10-21
Error: Invalid tag "::php::fpm::pool"


bobweston:
I'm trying to get this module working (I'm currently using nodes-php in production, but would like to migrate, if I can get this working)

My hiera looks like this:

<pre>classes:
  - php
php::fpm::pools:
  www:
   listen: '/var/run/php5-fpm.sock' 
   listen_owner: interact
   listen_group: interact
   listen_mode: '0660'
   user: interact
   group: interact
php::settings:
     PHP/upload_max_filesize: '20M'
     PHP/max_file_uploads: '21'	
</pre>	
When I try to spin up and provision the vm (vagrant), I get the following error:

<pre>Error: Invalid tag "::php::fpm::pool" at /tmp/vagrant-puppet/modules-078fcdc8fc841e4b19c52c75025d7110/php/manifests/fpm.pp:51 on node xxxxxxx.org</pre>

That of course points to this

<pre>create_resources(::php::fpm::pool, $real_pools)</pre>

I've tried just invoking it in my manifest by doing include ::php. But that has the same result as setting the class in Hiera.

Any thoughts on what I'm doing wrong? 

I've been banging my head for a couple hours trying to get this working. I am appreciative of any and all help.

I should add that removing all my hiera bits and simply invoking the ::php class <pre>include ::php</pre> results in the same error.

So I'm reasonably certain it is not a problem with the hiera bits.

cdenneen:
Should fix it https://github.com/mayflower/puppet-php/pull/130 hopefully

fpletz:
Hmm, interesting. We never stumbled on that problem.

What version of puppet are you using?

cdenneen:
It's a legacy issue... 

puppet 3.2.3 breaks with:
```puppet
  $real_pools = hiera_hash('php::fpm::pools',  $pools)
  create_resources('::php::fpm::pool', $real_pools)
```
puppet 3.8.3 works fine with it.

Probably for interim I would keep the syntax as:
```puppet
  $real_pools = hiera_hash('php::fpm::pools',  $pools)
  create_resources('php::fpm::pool', $real_pools)
```
which seems to work on all versions of Puppet 3.x

globin:
3.2 is so outdated I'm inclined to say that we don't fix this here and instead encourage you to patch it up for your systems and upgrade as soon as possible..

@fpletz, @fadenb ?

cdenneen:
[...]

PR #130, opened 2015-10-21
Fixed syntax error

cdenneen:
```Error: Invalid tag "::php::fpm::pool" at /.../php/manifests/fpm.pp:51 on node ...```

### DIFF
--- a/manifests/fpm.pp
+++ b/manifests/fpm.pp
@@ -48,7 +48,7 @@ class php::fpm (
   anchor { '::php::fpm::end': }
 
   $real_pools = hiera_hash('php::fpm::pools',  $pools)
-  create_resources(::php::fpm::pool, $real_pools)
+  create_resources('php::fpm::pool', $real_pools)
 
   # Create an override to use a reload signal as trusty and utopic's
   # upstart version supports this

### CODE REVIEW

File: manifests/fpm.pp [hunk 1]
```diff
@@ -48,7 +48,7 @@
   anchor { '::php::fpm::end': }
 
   $real_pools = hiera_hash('php::fpm::pools',  $pools)
-  create_resources(::php::fpm::pool, $real_pools)
+  create_resources('php::fpm::pool', $real_pools)
```
fpletz: Does `'::php::fpm::pool'` also solve this problem for you? It's generally better and faster to use the fully qualified form to refer to classes and resources.

cdenneen:
```puppet
  $real_pools = hiera_hash('php::fpm::pools',  $pools)
  create_resources('::php::fpm::pool', $real_pools)
```
results in:
Error: Invalid tag "::php::fpm::pool"

```puppet
  $real_pools = hiera_hash('php::fpm::pools',  $pools)
  create_resources('php::fpm::pool', $real_pools)
```
results in:
no error

globin: Which puppet version are you on?

cdenneen: Some legacy and some current (3.2.3 and 3.8.3)

fpletz:
This is a problem with Puppet 3.2.x. See #129.

Issue #78, opened 2015-04-29
php settings are misplaced on RHEL7


vchepkov:
I assume php::settings hiera should populate /etc/php.ini

```
php::settings:
  PHP/expose_php: Off
```

Instead it creates file /etc/php-cli.ini, which should be managed by php::cli::settings

I have PE 3.3 on RHEL7

fpletz:
Thanks for reporting this. We definitely have a bug here for RedHat-based systems concering the location of the configuration files.

On a CentOS 7 VM:

```
$ php -v
PHP 5.4.16 (cli) (built: Oct 31 2014 12:59:36) 
Copyright (c) 1997-2013 The PHP Group
Zend Engine v2.4.0, Copyright (c) 1998-2013 Zend Technologies
```

```
$ php --ini
Configuration File (php.ini) Path: /etc
Loaded Configuration File:         /etc/php.ini
Scan for additional .ini files in: /etc/php.d
Additional .ini files parsed:      /etc/php.d/curl.ini,
/etc/php.d/fileinfo.ini,
/etc/php.d/json.ini,
/etc/php.d/phar.ini,
/etc/php.d/zip.ini
```

```
$ php-fpm -i
[..]
Configuration File (php.ini) Path => /etc
Loaded Configuration File => /etc/php.ini
Scan this dir for additional .ini files => /etc/php.d
Additional .ini files parsed => /etc/php.d/curl.ini,
/etc/php.d/fileinfo.ini,
/etc/php.d/json.ini,
/etc/php.d/phar.ini,
/etc/php.d/zip.ini
[..]
```

So both the cli and fpm share the same `php.ini` which is `/etc/php.ini`. Seems a bit weird coming from Debian/Ubuntu.

I think that on RedHat-based systems we should fail when trying to use `php::cli::settings` and `php::fpm::settings`, and set `php::settings` in `/etc/php.ini`. @globin Your opinion?

globin:
Or maybe merge `php::cli::settings` and `php::fpm::settings` to `php::settings` and add a warning?

cprior:
I am /very/ new to Puppet but my problem might be related:
- followed instructions at http://stdout.no/a-modern-puppet-master-from-scratch/
-  Puppetfile contains amongst other things:
classes:
  - php:
      ensure: latest
      php::settings:
        Date/date.timezone: Europe/Berlin
- the puppet agent runs without error (- mysql-server \    ensure running works as expected)
[...]

PR #132, opened 2015-11-17
Separate php.ini settings from fpm settings

minorOffense:
Alright, so this is going to need some testing. I added a new variable to store the location of php.ini for each OS. Everything except Redhat was an "educated guess" as to the proper location.

The new "php::global" class runs in the init.pp file and sets values in the file (similarly to php-cli.ini)

The fpm settings have been moved into a php-fpm.ini file which according to the PHP docs should be autoloaded based on the package name (exactly the same way php-cli.ini works).

I provisioned a machine and tested the config. Appears to work and nothing breaks. But like I said, I haven't been able to test the other OSes yet.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -76,6 +76,13 @@ class php (
     } ->
   anchor { 'php::end': }
 
+  # Configure global PHP settings in php.ini
+  Anchor['php::begin'] ->
+  class {'::php::global':
+    settings => $real_settings,
+  } ->
+  anchor { 'php::end': }
+
   if $fpm {
     Anchor['php::begin'] ->
       class { '::php::fpm':
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -81,7 +81,7 @@ class php (
   class {'::php::global':
     settings => $real_settings,
   } ->
-  Anchor ['php::end':]
+  Anchor ['php::end']
 
   if $fpm {
     Anchor['php::begin'] ->
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -81,7 +81,7 @@ class php (
   class {'::php::global':
     settings => $real_settings,
   } ->
-  anchor { 'php::end': }
+  Anchor ['php::end':]
 
   if $fpm {
     Anchor['php::begin'] ->
--- a/manifests/global.pp
+++ b/manifests/global.pp
@@ -1,0 +1,33 @@
+# Install and configure mod_php for fpm
+#
+# === Parameters
+#
+# [*inifile*]
+#   Absolute path to the global php.ini file. Defaults
+#   to the OS specific default location as defined in params.
+# [*settings*]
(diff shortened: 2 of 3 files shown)


minorOffense:
Working on https://github.com/mayflower/puppet-php/issues/78

minorOffense:
Well there's work to do. Any hints you might be able to toss my way to figure out why the tests broke would be great.

fpletz:
Looking good after a quick glance if the behavior is indeed as stated in your link in #78. Will test this in a few days and get back to you.

We don't currently have a good test coverage. As the `php::config` resource already has specs, I think we can skip those for your new class.

minorOffense:
Doing some of my own testing, I noticed a warning on one of our vms

``
==> default: Could not retrieve fact='phpversion', resolution='<anonymous>': undefined method `split' for nil:NilClass
``

Not sure if it's related or not. We've been messing with installing newer versions of PHP on centos. But if you run into this too then odds are it's due to the patch.

minorOffense:
Just checking in to see if you need anything from me. Want to help with the patch as much as possible.

codec:
@minorOffense This is thanks to my :hankey: code in `lib/facter/phpversion.rb`. It's probably only happening on the first run?

minorOffense:
Yeah it's only on the first run. Ok so that's not related. Good to know ;-)

globin:
Docs say that it will always fail on first run.
https://github.com/mayflower/puppet-php#facts

manifests/init.pp
```diff
@@ -76,6 +76,13 @@
     } ->
   anchor { 'php::end': }
 
+  # Configure global PHP settings in php.ini
+  Anchor['php::begin'] ->
+  class {'::php::global':
+    settings => $real_settings,
+  } ->
+  Anchor ['php::end']
```
codec: @minorOffense The whitespace is causing the specs to fail using future parser. :see_no_evil:

mmoll:
@minorOffense could you rebase onto current master?

minorOffense:
Merged in changes from master but didn't fix that whitespace thing yet.

fpletz:
Squashed, rebased and pushed to master with the whitespace fix. Thanks!

voxel01:
The fix breaks backward compatibitilty. As soon as you use php::settings in hiera you get an compileerror:
```Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Duplicate declaration: Php::Config::Setting[/etc/php.ini: Date/date.timezone] is already declared; cannot redeclare at /etc/puppet/modules/common/php/manifests/config.pp:32 on node client.puppet.local```

The error was raised when I added the follwing config to hiera file. Which was correct until now.
```
php::settings:
    "Date/date.timezone": "Europe/Vienna"
```

mmoll:
4.0.0 is a major release because of all the backward-incompatibilities.

mmoll:
if the docs need updating, please open a PR :)

PR #137, opened 2015-11-28
removing any reference of hiera and replacing it with Puppet DSL code

seanscottking:
This PR resolves https://github.com/mayflower/puppet-php/issues/118. In a nutshell: this module used to depend heavily on Hiera and broke some best practice rules for puppet module design. Much of the groundwork to remove the Hiera dependency has already been done with previous PRs, and I now propose this PR to finally remove any `hiera` or `hiera_hash` lookups from the manifest code, and to make better documentation for proper example usage of Puppet's DSL code.

It includes:
  1. Updated `README.md`: Removed references of hiera data examples and replaced them with examples of Puppet DSL code.
  1. Updated `README.md`: Rearranged/reworded a few areas for better grammar, etc.
  1. Updated `README.md`: Given the Hiera-ish history of this module, I included a section for proper usage of Hiera given today's best practices.
  1. Updated manifests to not have `hiera` or `hiera_hash` lookups.

Please double check my work and let me know if it needs any adjustments, thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,106 +3,177 @@
 [![GitHub version](https://badge.fury.io/gh/mayflower%2Fpuppet-php.svg)](https://github.com/mayflower/puppet-php)
 [![Build Status](https://travis-ci.org/mayflower/puppet-php.svg?branch=master)](https://travis-ci.org/mayflower/puppet-php)
 
-mayflower/php is a Puppet module for managing PHP with a strong focus
-on php-fpm. We strive to support all recent versions of Debian, Ubuntu,
-RedHat/CentOS, openSUSE/SLES and FreeBSD. Managing Apache with `mod_php`
-is not supported.
+mayflower/php is a Puppet module for managing PHP with a strong focus on php-fpm. The module aims to use sane defaults for the supported architectures. We strive to support all recent versions of Debian, Ubuntu, RedHat/CentOS, openSUSE/SLES and FreeBSD. Managing Apache with `mod_php` is not supported. 
 
 This originally was a fork of [jippi/puppet-php](https://github.com/jippi/puppet-php)
 (nodes-php on Puppet Forge) but has since been rewritten in large parts.
 
 ## Usage
 
-The module aims to use sane defaults for the supported architectures. You
-must use hiera to configure most aspects of this module in a simple way.
-
-The recommended way is to include the `php` main class in your manifests:
+Quickest way to get started is simply `include`'ing the _`php` class_.
 
 ```puppet
-include ::php
+include '::php'
 ```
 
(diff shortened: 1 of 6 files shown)


mlrobinson:
I love the direction of these changes, it's what we've been doing as well. Along with these changes, won't you need to also remove settings like this: https://github.com/mayflower/puppet-php/pull/137/files#diff-4392194377d12c90ae24fc1ab8b5980eL19

I'm getting those warnings for now, and just ignoring them, as I need to override some settings, but I don't want to use their hiera structure (we are trying to prefix all our profile work with `profiles::`).

globin:
I still don't see a value in these changes, and you can do this with the current code without any need for hiera.. Apart from that you are heavily breaking our code relying on the possibility of hash merging by removing the explicit hiera calls.. This has been discussed earlier in some other issue or pr..

I'd propose closing this without merging, @fpletz, @fadenb,

mlrobinson:
The normal impetus for this, is to remove all knowledge of hiera from modules, and do that type of lookup/merge at a different level, which we call profiles. This keeps your module from having to care if hiera ever goes away/gets a name change, and just puts the configs it needs in place.

For my use, the code as it is can be ignored and worked around (even if it is noisy with the warnings), but it does force anyone wanting to use your module to at least have hiera installed, even if they aren't using it for whatever reason. That makes it less portable. I plan on working around it either way, but just answering the question of "why" for now.

globin:
Please look for the pr or issue where I explained that lengthily because of bugs in hiera/puppet

mlrobinson:
That would be https://github.com/mayflower/puppet-php/issues/118, which is the one referenced in the original comment. 

I can see the need for keeping the deep merge, but calling hiera here is the real issue. You could just merge the two hashes without it, and tell people to either do dynamic binding, or explicit, when they go to pass in data to the module.

The code works as is, so I can live without it being changed. It just seems like something that could be fixed and not force people to use the hiera vars you define here. I would say to keep the README changes, if those are fine, since those just show how to use the module without Hiera.

globin:
Using parameter lookup with deep merging does not work

seanscottking:
At this time I don't yet understand the above conversation or the problem with me removing the hiera lookups, so I can't promise that I will make a new PR attempt for that any time soon. But I will keep it in mind.

I will create a new PR with the `README.md` changes only.

globin:
Example:
common.yaml
```
php::extensions:
  intl: {}
```

devhost.yaml
```
php::extensions:
  xdebug: {}
```

This gets merged when using hiera_hash but not with param binding. We rely on this in nearly every of our projects.

PR #139, opened 2015-12-03
Add options to fpm config and pool configs

mlrobinson:
We manage a couple configuration options that aren't set in the current templates. Added the following options:

In php-fpm.conf, you can now modify process.max, and rlimit_files at this global level.

In the pool template, you can now modify the access log path and format.

### DIFF
--- a/manifests/fpm/config.pp
+++ b/manifests/fpm/config.pp
@@ -40,6 +40,12 @@
 # [*process_control_timeout*]
 #   The php-fpm process_control_timeout
 #
+# [*process_max*]
+#   The maximum number of processes FPM will fork.
+#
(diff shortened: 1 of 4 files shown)


globin:
Could you set the default for rlimit_files in the manifest to avoid the if in the template?

mlrobinson:
Gladly, as that's how I was going to set it before. I just wasn't sure what limit to set. If it's not set, it takes the system default. Setting a limit in the file seemed like it would override that without any good way of knowing what that default value should be per system.

fpletz:
Looks good and your reasoning for `rlimit_files` makes sense. Thanks! :beers:

PR #144, opened 2015-12-06
Fixes pecl provider

prachetasp:
This PR:
  - Discards header lines of the output of pecl list that are emphasized with coloration
  - Pecl does case-insensitive searches so downcases pecl package names to always match

Bold font in the shell output of `pecl list` caused the regex checks to fail. In addition, `pecl list` capitalized some values causing the `==` check to fail. To fix we downcase. This was tested with APC on CentOS.


fpletz:
Looking good, thanks for fixing this! :beers:

PR #145, opened 2015-12-10
add details of PPA override

brahman81:
Makes sense to have this in the README as I had to dig in source code to find how to specify an alternate PPA, this was to use PHP 5.6.


globin:
Thanks :+1:

PR #141, opened 2015-12-03
Updated README.md with better examples

seanscottking:
For some history read this issue: https://github.com/mayflower/puppet-php/issues/118

And then read the comments in this PR that got closed without merge: https://github.com/mayflower/puppet-php/pull/137

PR #147, opened 2015-12-14
Fixes non-overridable paths

### LINKED ISSUE #146: PHP7 Support
Hi, 

I'm trying to configure PHP7 with this module (we're already using it with PHP5), and I'm seeing a few places that hardcode in php5. Is there any plan to adjust these settings so it will work correctly with PHP7?  

Thanks!

cgalvarez:
When I have tried to install the recently released PHP 7.0.0, I have come across some issues:

1. The path for FPM pool conf files are built with `$::php::params::fpm_pool_dir` instead of `$::php::fpm::config::pool_base_dir`. This makes impossible to override the path of these files, but this path can be overriden with previous parameter from class `php::fpm::config`, and that path should be used because the pools belong to FPM and should use its base dir.
2. The path for PHP extensions ini files are built with `$::php::params::config_root_ini`, and we have the same previous problem: it is impossible to override that path. This should be a parameter from the defined type `php::extension`, but it would not be *hierable* due to the intrinsic nature of the defined type. The best approach I have come along is adding a new parameter `$::php::config_root_ini` to class `php` (at `manifests/init.pp`), so it becomes tuneable through manifests and Hiera also.
3. `php5enmod` and `php5query` commands in debian/ubuntu becomes `phpenmod` and `phpquery` with PHP 7, so the module fails to enable the requested extensions in those systems. The path to those tools could be added as parameter of the defined type `php::extension`, but, as it is a global command for the machine, I think it should be in the main class `php` at `init.pp`, where I have added them. I have added their defaults in `php::params`, so they inherit properly the default values when not provided.
4. All paths can be easily changed if a new param `$::php::params::cfg_root` is added to `php::params`.

So firstly I used the following Hiera setup to get PHP 7 working in my ubuntu virtual machine:

```
php::cli::inifile: '/etc/php/7.0/cli/php.ini'
php::extensions:
  'xdebug':
    ensure: '2.4.0RC3'
    provider: 'pecl'
php::fpm::config::config_file: '/etc/php/7.0/fpm/php-fpm.conf'
php::fpm::config::error_log: '/var/log/php7.0-fpm.log'
php::fpm::config::inifile: '/etc/php/7.0/fpm/php.ini'
php::fpm::config::pid_file: '/var/run/php/php7.0-fpm.pid'
php::fpm::config::pool_base_dir: '/etc/php/7.0/fpm/pool.d'
php::fpm::inifile: '/etc/php/7.0/fpm/php.ini'
php::fpm::service::service_name: 'php7.0-fpm'
php::fpm::pools:
  'www':
    listen: '/var/run/php/php7.0-fpm.sock'
php::package_prefix: 'php7.0-'
php::repo::ubuntu::ppa: 'ondrej/php-7.0'
```

But then:

* **1** forced pool to be at '/etc/php5/fpm/pool.d/www', but it should be at '/etc/php/7.0/fpm/pool.d/www'. After applying, the pull request fixes issue **1** it and the module works as expected.
* **2** forced `xdebug` extension conf file to be at `'/etc/php5/mods-available/xdebug.ini'`, but it should be at `'etc/php/7.0/mods-available/xdebug.ini'`. After applying, the pull request fixes issue **2** and the module works as expected.
* **3** fired puppet error `Could not find command '/usr/sbin/php5query'` or `Could not find command '/usr/sbin/php5enmod'`. After applying, the pull request fixes **3** and the module works as expected.
* after applying the pull request, hiera setup becomes easier and shorter.

After applying the pull request, the Hiera code becomes:

```
php::extensions:
  'xdebug':
    ensure: '2.4.0RC3'
    provider: 'pecl'
php::fpm::config::error_log: '/var/log/php7.0-fpm.log'
php::fpm::config::pid_file: '/var/run/php/php7.0-fpm.pid'
php::fpm::service::service_name: 'php7.0-fpm'
php::fpm::pools:
  'www':
    listen: '/var/run/php/php7.0-fpm.sock'
php::package_prefix: 'php7.0-'
php::params::config_root: '/etc/php/7.0'
php::repo::ubuntu::ppa: 'ondrej/php-7.0'
```

I have only tested this in Ubuntu 14.04, but the changes proposed in the pull request should not introduce errors in any other system. Finally, the pull request makes the module much more customizable.

Resolves #146

### DIFF
--- a/manifests/fpm/pool.pp
+++ b/manifests/fpm/pool.pp
@@ -98,6 +98,11 @@
 # [*php_directives*]
 #   List of custom directives that are appended to the pool config
 #
+# [*base_dir*]
+#   The folder that contains the php-fpm pool configs. This defaults to a 
+#   sensible default depending on your operating system, like
+#   '/etc/php5/fpm/pool.d' or '/etc/php-fpm.d'
+#
 define php::fpm::pool (
   $ensure = 'present',
   $listen = '127.0.0.1:9000',
@@ -138,9 +143,14 @@ define php::fpm::pool (
   $php_admin_flag = {},
   $php_directives = [],
   $root_group = $::php::params::root_group,
+  $base_dir = undef,
 ) {
 
   include ::php::params
+
+  if $base_dir != undef {
+    validate_absolute_path($base_dir)
+  }
 
   $pool = $title
 
@@ -154,13 +164,14 @@ define php::fpm::pool (
     default   => $::php::fpm::package,
   }
 
+  $pool_base_dir = pick_default($base_dir, $::php::fpm::config::pool_base_dir, $::php::params::fpm_pool_dir)
   if ($ensure == 'absent') {
-    file { "${::php::params::fpm_pool_dir}/${pool}.conf":
+    file { "${pool_base_dir}/${pool}.conf":
       ensure => absent,
       notify => Class['::php::fpm::service'],
     }
   } else {
-    file { "${::php::params::fpm_pool_dir}/${pool}.conf":
+    file { "${pool_base_dir}/${pool}.conf":
       ensure  => file,
       notify  => Class['::php::fpm::service'],
       require => Package[$real_package],
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -168,18 +168,21 @@ define php::extension(
     }
   }
 
+  $config_root_ini = pick_default($::php::config_root_ini, $::php::params::config_root_ini)
   ::php::config { $title:
-    file   => "${::php::params::config_root_ini}/${lowercase_title}.ini",
+    file   => "${config_root_ini}/${lowercase_title}.ini",
     config => $final_settings,
   }
 
   # Ubuntu/Debian systems use the mods-available folder. We need to enable
   # settings files ourselves with php5enmod command.
+  $ext_tool_enable = pick_default($::php::ext_tool_enable, $::php::params::ext_tool_enable)
+  $ext_tool_query  = pick_default($::php::ext_tool_query, $::php::params::ext_tool_query)
   if $::osfamily == 'Debian' {
-    $cmd = "/usr/sbin/php5enmod ${lowercase_title}"
+    $cmd = "${ext_tool_enable} ${lowercase_title}"
 
     exec { $cmd:
-      unless  => "/usr/sbin/php5query -s cli -m ${lowercase_title}",
+      unless  => "${ext_tool_query} -s cli -m ${lowercase_title}",
       require =>::Php::Config[$title],
     }
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -33,6 +33,19 @@
 #   to a sensible default depending on your operating system, like 'php-' or
 #   'php5-'.
 #
+# [*config_root_ini*]
+#   This is the path to the config .ini files of the extensions. This defaults
+#   to a sensible default depending on your operating system, like
+#   '/etc/php5/mods-available' or '/etc/php5/conf.d'.
+#
+# [*$ext_tool_enable*]
+#   Absolute path to php tool for enabling extensions in debian/ubuntu systems.
+#   This defaults to '/usr/sbin/php5enmod'.
+#
+# [*$ext_tool_query*]
+#   Absolute path to php tool for querying information about extensions in
+#   debian/ubuntu systems. This defaults to '/usr/sbin/php5query'.
+#
 class php (
   $ensure         = $::php::params::ensure,
   $manage_repos   = $::php::params::manage_repos,
@@ -46,6 +59,9 @@ class php (
   $extensions     = {},
   $settings       = {},
   $package_prefix = $::php::params::package_prefix,
+  $config_root_ini = $::php::params::config_root_ini,
+  $ext_tool_enable = $::php::params::ext_tool_enable,
+  $ext_tool_query  = $::php::params::ext_tool_query,
 ) inherits ::php::params {
 
   validate_string($ensure)
@@ -59,6 +75,15 @@ class php (
   validate_bool($phpunit)
   validate_hash($extensions)
   validate_hash($settings)
+  if $config_root_ini != undef {
+    validate_absolute_path($config_root_ini)
+  }
+  if $ext_tool_enable != undef {
+    validate_absolute_path($ext_tool_enable)
+  }
+  if $ext_tool_query != undef {
+    validate_absolute_path($ext_tool_query)
+  }
 
   # Deep merge global php settings
   $real_settings = deep_merge($settings, hiera_hash('php::settings', {}))
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,6 +1,12 @@
 # PHP params class
 #
-class php::params {
+class php::params(
+  $cfg_root = undef,
+) {
+
+  if $cfg_root != undef {
+    validate_absolute_path($cfg_root)
+  }
 
   $ensure              = 'present'
   $fpm_service_enable  = true
@@ -15,8 +21,8 @@ class php::params {
 
   case $::osfamily {
     'Debian': {
-      $config_root             = '/etc/php5'
-      $config_root_ini         = "${::php::params::config_root}/mods-available"
+      $config_root             = pick($cfg_root, '/etc/php5')
+      $config_root_ini         = "${config_root}/mods-available"
       $common_package_names    = []
       $common_package_suffixes = ['cli', 'common']
       $cli_inifile             = "${config_root}/cli/php.ini"
@@ -35,6 +41,8 @@ class php::params {
       $package_prefix          = 'php5-'
       $compiler_packages       = 'build-essential'
       $root_group              = 'root'
+      $ext_tool_enable         = '/usr/sbin/php5enmod'
+      $ext_tool_query          = '/usr/sbin/php5query'
 
       case $::operatingsystem {
         'Debian': {
@@ -52,7 +60,7 @@ class php::params {
     }
 
     'Suse': {
-      $config_root             = '/etc/php5'
+      $config_root             = pick($cfg_root, '/etc/php5')
       $config_root_ini         = "${config_root}/conf.d"
       $common_package_names    = ['php5']
       $common_package_suffixes = []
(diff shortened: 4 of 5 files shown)


globin:
Tests are failing but looks good otherwise :wink:

cgalvarez:
I've tested it only locally in my ubuntu vagrant virtual machine, but I will take a look at those tests to guess what is failing and make the pull request fully compliant :wink:

cgalvarez:
Hi, globin. Now 7 tests pass, but 3 still fail.

By the way, the main problem with the tests was that the defined type `$::php::fpm::pool` was using the input parameter `$::php::fpm::pool::base_dir` to build the path to the pool conf file. This same problem will occur with other parameters like `$::php::fpm::pool::user`, `$::php::fpm::pool::root` or `$::php::fpm::pool::group_root`, because they inherit from class `$::php::params`, but that class is not available when instantiating a resource of that defined type directly from a manifest.

That was the problem with my approach: I was using the variable `$::php::fpm::config::pool_base_dir`, but that variable is only accessible if you have previously instantiated the class `$::php::fpm::config`, so I have allowed providing the path through an input parameter, if not being provided, then it is inherited from `$::php::fpm::config` if it has been instantiated, and otherwise from `$::php::params` (included at the very beginning of the type definition).

Later I'll take a look on those three failing tests.

mmoll:
I'd recommend to use stdlib's `pick()` instead of the ternaries and to squash all the commits together into one.
:+1: otherwise

cubiclelord:
Any particular reason why you're running the xdebug-2.4.0RC3 from PECL instead of 2.4.0RC2 from ondrej's PHP7 packages?

cgalvarez:
I've fixed another misbehaviour in defined type `php::extension`. It created the ini file `php::config` using the path `$::php::params::config_root_ini`, but that path now should be overriden by `$::php::config_root_ini` when the related class `php` has been instantiated.

@cubiclelord Since they are not final stable releases, the latest release candidate fixes bugs and improves performance comparing with the previous ones. In particular, from the announcements in the [xdebug homepage](http://xdebug.org/) the statement that made me choose 2.4.0RC3 was *This release candidate fixes weird optimisation issues with GCC 4.8*. Sincerely I don't know if it affects me someway, but my general preference is choosing the latest non-final release until the final stable one from same branch is released.

@mmoll Thanks for the suggestions. I agree. Ternaries in `php::params` changed. Commits squashed (it's my first squash, so if I've done something wrong, please, tell me).

All checks fail now. I took a look at `/spec/defines/extension_spec.rb`, and a variable `etcdir` is hardcoded there for Debian systems pointing to `/etc/php5/mods-available`, which really depends on the desired version of PHP.

Besides the above, when I run 'bundle exec rake', it stops after making the lint subtask. Any ideas on what's happening? I think this is the reason for all Travis CI fails. Looking into stackoverflow, but still nothing.

Any other suggestions or warnings about detected bugs/misbehaviours are welcome :wink:

mmoll:
`manifests/extension.pp - WARNING: line has more than 80 characters on line 89` - I think we should bump this to 120... @fpletz? @globin?

mmoll:
@cgalvarez please use LF (UNIX style)

mmoll:
Please rebase onto current master, the 80chars limit was removed.

cgalvarez:
@mmoll I **do** use LF :wink:... don't have a clue on how the LF/CR get there (suppose that one of the commands when I was learning how to squash commits did the joke).

All LF/CR removed, now all tests pass (cannot use `pick` in the ternaries in `manifests/extension.pp` because they make tests fail). A hardcoded path in `spec/defines/extension_rspec.rb` fixed. Rebase into master done.

mmoll:
@cgalvarez almost there... :)

does the following work as expected for you? At least tests pass with it...

```diff
diff --git a/manifests/extension.pp b/manifests/extension.pp
index fbfc4c7..906a83e 100644
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -168,19 +168,9 @@ define php::extension(
     }
   }
 
-  # Approach #1
-  $config_root_ini = $::php::config_root_ini ? {
-    undef   => $::php::params::config_root_ini,
-    default => $::php::config_root_ini,
-  }
-  $ext_tool_enable = $::php::ext_tool_enable ? {
-    undef   => $::php::params::ext_tool_enable,
-    default => $::php::ext_tool_enable,
-  }
-  $ext_tool_query = $::php::ext_tool_query ? {
-    undef   => $::php::params::ext_tool_query,
-    default => $::php::ext_tool_query,
-  }
+  $config_root_ini = pick_default($::php::config_root_ini, $::php::params::config_root_ini)
+  $ext_tool_enable = pick_default($::php::ext_tool_enable, $::php::params::ext_tool_enable)
+  $ext_tool_query = pick_default($::php::ext_tool_query, $::php::params::ext_tool_query)
 
   ::php::config { $title:
     file   => "${config_root_ini}/${lowercase_title}.ini",
```

manifests/fpm/pool.pp
```diff
@@ -154,13 +164,22 @@
     default   => $::php::fpm::package,
   }
 
+  if $base_dir != undef {
+    $pool_base_dir = $base_dir
+  } else {
+    $pool_base_dir = tagged(php::fpm::config) ? {
```
mmoll: tagged?

cgalvarez:
@mmoll Didn't know about function `pick_default`. I've changed both `manifests/fpm/pool.pp` and `manifests/extension.pp`. Both works as expected and all tests pass.

The only problem now is that automatic tests will stop at rake lint task, because of current setting `config.fail_on_warnings = true` in `Rakefile`. You said before that the 80chars limit was removed, but I can't see that in the latest version of the `Rakefile` in the master branch of your repo. Should I add the related `PuppetLint.configuration.send('disable_80chars')` line?

Waiting for your response to rebase again.

mmoll:
@cgalvarez see f246596, the config was moved to a sperate file

mmoll:
maybe condense the commit message and use `closes GH-146` for auto-close of that issue, as 147 is this PR which will be closed when merging anyway.
:+1: otherwise.

mmoll:
@fpletz @globin I think this is good to go :+1:

tobru:
:+1: from my side for this PR to get merged now

cgalvarez:
I just picked a bug. `phpenmod` executes always when applying multiple consecutive times.

`phpquery` now requires a **third** parameter to be passed: the PHP version. I cannot find information about neither `phpenmod` nor `phpquery`. `man` does not have any info about them.

After try/error and the error messages printed, the version to provide must be `7.0` (the PHP version), so the `unless` command becomes `phpquery -s cli -m ext_name -v 7.0`. I think we could use the php version fact to build the command, but I'm stuck since cannot check successfully that the extension is installed with `phpquery` (I'm trying with the package `php7.0-mysql`). The following commands always return the message `No module matches xxxxx` (where `xxxxx` is the module name provided with `-m`):

- `phpquery -s cli -m mysql -v 7.0`
- `phpquery -s cli -m php7.0-mysql -v 7.0`

The module is loading, because my WordPress installation reads the database without problems, so I think that the module is being registered successfully by the current setup, but it is always executed since the `unless` parameter is failing.

Any help on this?

cubiclelord:
In my hacked up version I ended up using a php_major_version variable which defaulted to 5 but which could be overwritten in  hiera to be 7.  Then I could easily know which commands needed to be run.

cgalvarez:
@cubiclelord I just realized that the required parameters for both commands can be included in the already created parameters `$::php::ext_tool_enable` and `$::php::ext_tool_query`, so I think there is no need to change that.

Just simply pass `php::ext_tool_query: '/usr/sbin/phpquery -v 7.0'`. The same for `phpenmod`.

The folder for PHP7 extensions must be `php::config_root_ini: '/etc/php/mods-available'` (my mistake at the first comment).

PHP7 comes bundled with `json`, `pdo_mysql`, `mysqli` and `opcache`. If I run `phpquery -s cli -v 7.0 -m mysqli`, it prints out `mysqli (Enabled for cli by maintainer script)`. This is so because there is an empty file at `/var/lib/php/modules/7.0/cli/enabled_by_maint/mysqli`. Nevertheless, it doesn't matter if such file exists or not (this is, if the module is registered or not), because if you execute `phpquery` for `mysqli` and another arbitrary string like `blablabla`, you always get exitcode 0. Being this so, the `unless` parameter with `phpquery` becomes unuseful for PHP7.

There is another folder `/var/lib/php/modules/registry` which gets filled up with the registered modules (I can see the files for `xdebug` and `mysql`, installed through `pecl` and `apt-get` respectively), so that is working fine, but I'm not able yet to register successfully xdebug/mysql modules (I've created a `test.php` file with content `<?php phpinfo(); ?>`, and cannot see the installed modules there).

Besides, the modules ini files won't get symlinked or created at the right folder (from `/etc/php/mods-available/` to `/etc/php/7.0/fpm/conf.d/`).

I think this is a bug in the `phpenmod` and `phpquery` tools, and not in the current PR.

cubiclelord:
I think I found an issue with $config_root_ini. When using the ondrej-php7 packages on Ubuntu 14.04, the extension configs are all stored in /etc/php/mods-avaialble, not /etc/php/7.0/mods-available, which is where the current module expects them to be. Any ideas about how to adjust this?

cgalvarez:
@cubiclelord Use the parameter `$::php::config_root_ini` to override it :wink:

cubiclelord:
Doh! Should have noticed that... :)

cgalvarez:
Another issue: XDebug 2.4.0RC3 should be loaded as a zend extension (`zend_extension=xdebug.so`), but when installed as a package from [Ondřej Surý PHP7 repo](https://launchpad.net/~ondrej/+archive/ubuntu/php-7.0) (@cubiclelord you got it updated some hours ago :wink:) it's loaded as a normal extension (`extension=xdebug.so`).

The parameter `$::php::extension::zend` could do the job, but it aborts the provisioning if provider is other than `pecl`.

BTW, XDebug installs fine from previous repo, but the package prefix must be set to `'php-'`.

tobru:
There are still some troubles when not using Hiera to set parameters. We should be able to f.e. configure PHP FPM paths in init.pp as parameter...

mmoll:
@tobru please comment in #141 regarding non-hiera usage.

mmoll:
@codec please rebase php-version-select :boom:

cubiclelord:
I would like to note for the record, to overwrite the config_root, you have to set:

php::params::cfg_root: '/etc/php/7.0'

NOT 

php::params::config_root: '/etc/php/7.0'

I am an idiot and stared at this for 30 minutes before realizing I had the variable set wrong.  I'm posting here, just in case it helps someone else.

mbrodala:
@cubiclelord Thanks for the heads up. @cgalvarez could you edit your OP to fix this? A lot of people will start by copying your code and fail until they read all comments here. ;-)

Issue #131, opened 2015-11-17
php 5.6 on Centos 7.1.1503


ajasak:
I'm trying to install php 5.6 on Centos 7.1.1503 by adding manage_repos => true.
It ends with error:
```
Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install php-common' returned 1: Error: Package: php-pecl-zip-1.13.1-1.el7.remi.5.6.x86_64 (remi-php56)
           Requires: libzip.so.4()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: /Stage[main]/Php::Packages/Package[php-common]/ensure: change from absent to present failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install php-common' returned 1: Error: Package: php-pecl-zip-1.13.1-1.el7.remi.5.6.x86_64 (remi-php56)
           Requires: libzip.so.4()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: Execution of '/usr/bin/yum -d 0 -e 0 -y install php-cli' returned 1: Error: Package: php-pecl-zip-1.13.1-1.el7.remi.5.6.x86_64 (remi-php56)
           Requires: libzip.so.4()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
Error: /Stage[main]/Php::Packages/Package[php-cli]/ensure: change from absent to present failed: Execution of '/usr/bin/yum -d 0 -e 0 -y install php-cli' returned 1: Error: Package: php-pecl-zip-1.13.1-1.el7.remi.5.6.x86_64 (remi-php56)
           Requires: libzip.so.4()(64bit)
 You could try using --skip-broken to work around the problem
 You could try running: rpm -Va --nofiles --nodigest
```
Solution for this error is to add remi or remi-safe repository to manifests/repo/redhat.pp:
```
contain "::yum::repo::remi"
```
or
```
yum::managed_yumrepo { 'remi-safe':
    descr         => 'Les RPM de remi pour Enterpise Linux $releasever - $basearch - PHP 5.6',
    mirrorlist    => 'http://rpms.famillecollet.com/enterprise/$releasever/safe/mirror',
    enabled       => 1,
    gpgcheck      => 1,
    gpgkey        => 'file:///etc/pki/rpm-gpg/RPM-GPG-KEY-remi',
[...]

PR #148, opened 2015-12-14
Fix for  php 5.6 on Centos 7.1.1503 #131 

heavyengineer:
have added the yum repo so this works ok on centos7 - not sure if i should leave the original reference in so have left it.

### DIFF
--- a/manifests/repo/redhat.pp
+++ b/manifests/repo/redhat.pp
@@ -10,4 +10,5 @@ class php::repo::redhat (
   $yum_repo = 'remi_php56',
 ) {
   contain "::yum::repo::${yum_repo}"
+  contain "::yum::repo::remi"
 }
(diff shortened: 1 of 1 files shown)


heavyengineer:
I am unable to see what failed 8(  Will fix it if i can see it. the module works fine for me locally tho so i assume maybe a test needs to be changed somewhere?

manifests/repo/redhat.pp
```diff
@@ -10,4 +10,5 @@
   $yum_repo = 'remi_php56',
 ) {
   contain "::yum::repo::${yum_repo}"
+  contain "::yum::repo::remi"
```
fpletz: We've put `puppet-lint` into full pedantic mode so that the code quality metric on the Puppet Forge is always at 100%. :panda_face: 

This is the relevant warning that is treated as an error:
```
manifests/repo/redhat.pp - WARNING: double quoted string containing no variables on line 13
```

Sorry, you'll have to use single quotes here. :)

fpletz:
Thanks!

PR #153, opened 2016-01-11
added condition to pear.pp for Amazon linux to prevent error if packa…

ryno75:
…ge_prefix is specified


codec:
Thanks for the contribution! :+1:

PR #155, opened 2016-02-08
Make it possible to enable/disable the fpm service.

andrekeller:
We use supervisor to manage individual fpm services, but still want the php module to install php-fpm. Currently its not possible to ensure that the service is disabled and stopped.


globin:
Thanks :+1:

PR #159, opened 2016-02-18
Allowed setting the FPM user from the root class

tombuildsstuff:
Given we'd rather not override this via specific parameters - we're making it possible to pass in the user/group on the root class


mmoll:
merged, thanks @tombuildsstuff!

Issue #161, opened 2016-02-23
overriding fpm_service_name has no effect


okolesnyk:
Due to PHP 7 upgrade, I tried the following to announce the new service name to the module NOT VIA HIERA.

I've got my own class with custom configuration for php and I need to override fpm_service_name and I've got no clue how to do this. 

I've tried something like this:

`    class { 'php::fpm::service':
        service_name => 'php7.0-fpm'
    }`

and I have something like this always:
Error: Duplicate declaration: Class[Php::Params] is already declared; cannot redeclare
or
Error: Duplicate declaration: Class[Php::Fpm::Service] is already declared; cannot redeclare

mmoll:
@okolesnyk You don't state how you are specifing the name, but it sounds like something similar to https://github.com/eri-ca/puppet-php/commit/7e7ed2f91722a3b515d64ca1aec3bec0ce9bfc78 and refactoring variable usage (init vs. params) in fpm/service.pp is needed to solve this.

okolesnyk:
@mmoll Yes it is. I just need to dynamically declare php-fpm service from my own class with custom configuration.
Is it possible for you to create pull-request to origin repo? My requirement is to use mayflower/puppet-php repo.

PR #163, opened 2016-02-23
Enable Override of $fpm_service_name in case of using php7.0

okolesnyk:
Added possibility to override fpm service name in case of using php7.0

Resolves:  GH-161

### DIFF
--- a/manifests/fpm.pp
+++ b/manifests/fpm.pp
@@ -8,6 +8,10 @@
 # [*service_ensure*]
 #   Ensure FPM service is either 'running' or 'stopped'
 #
+# [*service_name*]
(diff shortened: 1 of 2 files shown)


mmoll:
`$name` is reserved, please see https://docs.puppetlabs.com/puppet/latest/reference/lang_reserved.html#reserved-variable-names
by reverting the service_name -> name change back, the build should get green

mmoll:
please squash your commits together into one and add `closes GH-161` to the commit message, looking good otherwise :+1:

mmoll:
merged, дякуємо / спасибо @okolesnyk!

PR #166, opened 2016-03-02
class uses 'pools' hiera key, not 'pool'

vchepkov:
either module has to be changed or the documentation.
I would change module instead, since it's not intuitive, but this would create an incompatibility with current configurations

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ php::extensions:
     settings:
       'apc/stat': 1
       'apc/stat_ctime': 1
-php::fpm::pool:
+php::fpm::pools:
   www2:
     listen: '127.0.1.1:9000'
 ```


mmoll:
the corresponding class-syntax in line 108 also needs changing then.

vchepkov:
Not sure I follow.
I am pointing to this line of code:
https://github.com/mayflower/puppet-php/blob/master/manifests/fpm.pp#L77

it uses `hiera_hash('php::fpm::pools')`

mmoll:
@vchepkov oh, I see... that's correct, merged. That was a fallout of a recent readme change and the original code change was one of the steps to make this module more hiera-centric.
Спасибо @vchepkov!

vchepkov:
Bitte schön

PR #169, opened 2016-03-29
Enable/Disable $ext_tool Debian

SanderSander:
This fixes some issues with php7 and php5enmod/php5query on at least Ubuntu, but i assume it would also work on debian. 

Also switched to non deprecated php 7 repo.

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -176,9 +176,11 @@ define php::extension(
 
   # Ubuntu/Debian systems use the mods-available folder. We need to enable
   # settings files ourselves with php5enmod command.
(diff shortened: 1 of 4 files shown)


fpletz:
Thanks a lot!

With this commit osfamilies other than Debian will fail because `php::params::ext_tool_enabled` is not defined for them. Could you please default to `$ext_tool_enabled = false` in `params.pp`?

SanderSander:
Added the default to other families. 

Build will probably fail because of latest commit on the master https://github.com/mayflower/puppet-php/commit/c5ce1a5ad50608d55b3b50ad1ac1072bd976b00c

Not sure what to do with it, I can readd the $old_stable

fpletz:
Thanks! I'll merge this in this state. We have to stabilize master and release 4.0.0 soon.

PR #171, opened 2016-04-02
Add '/usr/local' paths to support FreeBSD

dennisdegreef:
FreeBSD installs all software from a package manager into `/usr/local`.
This includes the `wget` binary, which is included in the `pkg` package manager.

OSX will probably also benefit from this when `wget` is installed using Homebrew.


mmoll:
bedankt @dennisdegreef!

dennisdegreef:
Graag gedaan! :+1:

PR #179, opened 2016-04-13
Adding custom template option and options hash, fixing some formatting, aligning params

FutureSharks:
It would be great to be able to change the template used in the pool definition so I added this as an option that defaults to the included template file.

I also added an options hash which would also be handy when using a custom template.


fpletz:
Thanks!

PR #180, opened 2016-04-14
Adding params logic for php7 on Ubuntu/Debian

FutureSharks:
Here is my idea for PHP7 support for Ubuntu/Debian.

I couldn't think of a more elegant way of doing it. The difference in paths and naming between version 5 and 7 is not really consistent.

Tested on Ubuntu 16.04 with PHP7, 95% sure it will work on Debian stretch too. I don't have any other OS to test with here but perhaps the same approach could be followed?


manifests/params.pp
```diff
@@ -16,54 +29,73 @@
...
+
+      if $::lsbdistcodename == 'wheezy' {
+        $manage_repos = true
+      } elsif $::operatingsystem == 'Ubuntu' {
+        $manage_repos = true
+      } else {
+        $manage_repos = false
+      }
+
+      if $::lsbdistcodename == 'stretch' or $::lsbdistcodename == 'xenial' {
```
mwhahaha: I had to use $::os['distro']['codename'] == 'xenial' with php 4.4.2 as lsbdistcodename was undefined.

FutureSharks: Ah I see. Even though the lsbxxx facts are commonly used everywhere, they depend on the lsb-release package. We could change it to use the os hash but would need to retest.

manifests/params.pp
```diff
@@ -1,11 +1,24 @@
...
+# [*config_root_override*]
+#   The configuration root directory used in parameter calculation
+#
+# [*php_version_override*]
+#   The version of php used in parameter calculation
+#
 class php::params(
-  $cfg_root = undef, # lint:ignore:parameter_documentation
+  $config_root_override = undef,
+  $php_version_override = undef
```
fpletz: I'm not sure if we should add even more parameters to this class. The params class is inherited by the main `php` class and thus these parameters can't be overridden without Hiera. Quite a few of our users demanded that we make the module independent of Hiera. What do you think?

manifests/params.pp
fpletz: We're using the LSB facts in other places, too. I'd say keep this for now. I've opened a ticket to discuss & resolve this for the whole module: #184

manifests/params.pp
FutureSharks: I didn't think anyone used puppet without hiera but OK. Shall I just leave the params as they were? Or remove `cfg_root`, `cfg_root` and `php_version_override`?

mwhahaha: I'm using it without hiera, but as an option perhaps we could just allow for the parameters that consume these overrides to be defined somewhere else (like init or something)?

FutureSharks: OK how about I just move both to init.pp?

mwhahaha: I think that would be a logical place

FutureSharks:
OK I've updated the PR to move those two params to init.pp.

It seems a little strange to have parameters in init.pp that are used by params.pp to calculate some values that are used back in init.pp, but OK.

mmoll:
Just a side note: The pattern is quite normal for modules which are also friendly to non-hiera setups: calcuate the defaults in params.pp, pull these in as values init.pp, have it overridable there.

FutureSharks:
@mmoll I'm aware of that convention, but this is something a bit different I think? For example `override_php_version` can be set in init.pp, which is used by params.pp to calculate `config_root`, which is used to calculate `config_root_ini`, which is used back in init.pp.

mmoll:
@FutureSharks (un)fortunately this module was/is built "hiera centric", leading to different implementations...

FutureSharks:
Anyway, all good. Let me know if this PR works for you guys.

rfletcher:
I have the same problem to solve, so I was happy to find this PR, but I've run into an issue.

Hiera config:

````yaml
php::override_php_version: "7.0"
php::repo::ubuntu::ppa: "ondrej/php"
````

Puppet manifest:
````puppet
include ::php
````

Running Puppet (3.8.7), the `php::override_php_version` param has no effect, and I get these errors:

````
==> apps-web: Warning: Scope(Class[Php::Params]): Could not look up qualified variable
  '::php::override_php_version'; class ::php has not been evaluated
==> apps-web: Warning: Scope(Class[Php::Params]): Could not look up qualified variable
  '::php::override_config_root'; class ::php has not been evaluated
````

**Comment by FutureSharks** [comment]
@rfletcher you are right, it doesn't work. I don't think it can work due to the issue I was mentioning before you posted (recursion between init and params).

At this point I think we should either merge my initial version with parameters in params.pp or someone needs to do a larger rewrite.

**Comment by igalic** [comment]
how about the puppetlabs-postgresql aproach of a global class?

**Comment by FutureSharks** [comment]
Yeah that would work and is probably the correct way of doing it:
https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/params.pp#L2
https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/globals.pp

I'm not sure I have to write and test this approach right now but I could perhaps create a new PR next week.

**Comment by igalic** [comment]
@FutureSharks: do it so, :hash: :one:

**Comment by FutureSharks** [comment]
OK I've done a quick update and copied the arrangement with a `php::globals` class like in [puppetlabs-postgresql](https://github.com/puppetlabs/puppetlabs-postgresql). Should be pretty clear how to update the `php::globals` class for other OSs and PHP 7.

@rfletcher can you retest with these hiera settings?
```
php::globals::php_version: '7.0'
php::repo::ubuntu::ppa: 'ondrej/php'
```
For me this works on Ubuntu 14.04 and 16.04.

Travis says tests for Red Hat based distros are failing but I'm not really sure why. Any ideas?

**Comment by rfletcher** [comment]
> @rfletcher can you retest with these hiera settings?

Working for me, too, but there's another issue that I hadn't noticed before.

I wasn't using this module before your changes, so I can't say whether this issue is yours, but `phpenmod` is running with every puppet run, whether or not a given php module is already installed. I turned on debug puppet output and found a `phpquery` error saying that `-v` was required.

At least for my specific setup, doing something like this in extension.pp fixes it.

````diff
@@ -195,12 +196,12 @@ define php::extension(
+    include ::php::globals
 
     if $sapi == 'ALL' {
       exec { $cmd:
-        unless  => "${ext_tool_query} -s cli -m ${lowercase_title}",
+        unless  => "${ext_tool_query} -s cli -m ${lowercase_title} -v ${php::globals::globals_php_version}",
         require =>::Php::Config[$title],
       }
     } else {
       exec { $cmd:
-        unless  => "${ext_tool_query} -s ${sapi} -m ${lowercase_title}",
+        unless  => "${ext_tool_query} -s ${sapi} -m ${lowercase_title} -v ${php::globals::globals_php_version}",
         require =>::Php::Config[$title],
       }
     }
````

**Comment by FutureSharks** [comment]
@rfletcher 
It doesn't work on Ubuntu 16.04 anyway, see my note here:
https://github.com/mayflower/puppet-php/issues/146#issuecomment-217426525

Also, your suggested change won't work for PHP 5 because the `-v` option doesn't exist.

**Comment by FutureSharks** [comment]
@rfletcher check my last commit. I think that should fix it?

**Comment by rfletcher** [comment]
> @rfletcher check my last commit. I think that should fix it?

Yup, looks good to me. (Ubuntu 14.04, ondrej/php PPA, php 7.0)

**Comment by max-rocket-internet** [comment]
Can someone test on Red Hat or CentOS? Travis builds are failing but I can't see why.

**Comment by FutureSharks** [comment]
Can someone test on Red Hat or CentOS? Travis builds are failing but I can't see why.

File: manifests/params.pp [hunk 3]
```diff
@@ -98,7 +92,7 @@
       }
     }
     'RedHat': {
-      $config_root_ini         = '/etc/php.d'
+      $config_root             = $php::globals::globals_config_root
```
igalic: this is the reason rhel fails. `$config_root_ini` should be the same as `$config_root`, if i see this correctly

max-rocket-internet: You are right. This is my mistake. I'm just going to revert this part and leave it was as this PR is only for Debian/Ubuntu.

FutureSharks:
OK sorted, builds now pass. Can we merge this now?

tdm4:
I think this should be merged in because ondrej/php5-5.6 repo is deprecated. ondrej/php is the repo to use now.

mmoll:
I didn't test this, but as tests are green and there seem to be no objections I'll merge.

mmoll:
thanks!

PR #182, opened 2016-04-26
Allow specifying the sapi for debian/ubuntu systems

grigarr:
This patch enables specifying a specific sapi to the php5enmod executor. Some extensions (i.e. php5-suhosin-extension) are undesirable for the cli sapi, and this enables specifying only fpm.


mpdude:
Cool, have been waiting for this!

fpletz:
Thanks! :beers:

PR #189, opened 2016-06-02
allow for class to be defined and pass a pid file path

adamcousins:
allow of passing a pid file to globals like so:
```puppet
class { '::php::globals':
                php_version => '5.5',
                fpm_pid_file => '/var/run/php-fpm/php-fpm-5.5.pid',
        } ->
        class { '::php':
                ensure       => latest,
                manage_repos => false,
                fpm          => false,
                dev          => false,
                composer     => false,
                pear         => true,
                phpunit      => false,
        }
```
This was need to address amazon linux and a init.d pid file looking for:
`pidfile=${PIDFILE-/var/run/php-fpm/php-fpm-5.5.pid}`

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -8,6 +8,10 @@
 # [*config_root*]
 #   The configuration root directory.
 #
+# [*fpm_pid_file*]
+#   Path to pid file for fpm
+#
+
 class php::globals (
   $php_version = undef,
   $config_root = undef,
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -10,11 +10,10 @@
 #
 # [*fpm_pid_file*]
 #   Path to pid file for fpm
-#
 
 class php::globals (
-  $php_version = undef,
-  $config_root = undef,
+  $php_version  = undef,
+  $config_root  = undef,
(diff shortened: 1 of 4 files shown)


mmoll:
could you have a look at #180 and report back, if setting `fpm_pid_file` in `globals` would also solve this?

adamcousins:
@mmoll done!

manifests/globals.pp
```diff
@@ -8,9 +8,14 @@
 # [*config_root*]
 #   The configuration root directory.
 #
+# [*fpm_pid_file*]
+#   Path to pid file for fpm
+#
+
```
mmoll: superfluous newline

manifests/globals.pp
```diff
@@ -95,7 +104,7 @@
           }
           default: {
             $default_config_root = '/etc/php5'
-            $fpm_pid_file        = '/var/run/php5-fpm.pid'
+            $default_fpm_pid_file = '/var/run/php5-fpm.pid'
```
mmoll: indentation is now off here and in the changes above

mmoll:
merged, thanks @adamcousins!

Issue #191, opened 2016-06-08
Ordering issues with php::global applying before PHP packages installed


chrisboulton:
I've come across a strange ordering issue, where `php::global` seems to be included and attempting to apply resources (`Php::Global/Php::Config[global]/Php::Config::Setting[/etc/php5/php.ini: PHP/variables_order]`) before PHP is actually installed - at this point in time `/etc/php5/` doesn't exist, which causes a bunch of errors:

```
==> default: Error: Could not set 'present' on ensure: No such file or directory - /etc/php5/php.ini at 54:/var/tmp/vagrant-puppet/modules-c61fb460434b0ef3bc9fb916e18be6c3/php/manifests/config/setting.pp
==> default: Error: Could not set 'present' on ensure: No such file or directory - /etc/php5/php.ini at 54:/var/tmp/vagrant-puppet/modules-c61fb460434b0ef3bc9fb916e18be6c3/php/manifests/config/setting.pp
==> default: Wrapped exception:
==> default: No such file or directory - /etc/php5/php.ini
```

Down later in the Puppet run, the PHP packages are installed, and the directory structure is good. A second run succeeds.

A quick glance into init.pp seems to have this handled with the anchoring, so at this point in time I'm not sure why this is happening. I haven't dug further yet.

chrisboulton:
Actually, is the anchoring right?

```
  # Configure global PHP settings in php.ini
  Anchor['php::begin'] ->
  class {'::php::global':
    settings => $real_settings,
  } ->
  Anchor['php::end']
```

I'd have expected it to be the other way around - `php::end` must be complete before the settings (aaaand FPM down lower?) can be applied:

```
  # Configure global PHP settings in php.ini
  class {'::php::global':
    settings => $real_settings,
    require  => Anchor['php::end'],
  }
```

mmoll:
should be fixed by #194

PR #192, opened 2016-06-09
Don't manage global php.ini settings on Debian based systems

chrisboulton:
#132 resulted in a global php.ini being created/managed to fix some issues on RedHat based systems.

Debian (and derivatives) don't actually have an `/etc/php|php5/php.ini`. All configuration lives under a SAPI. Configuring these settings in this global file does nothing under this family of operating systems.

I came across this when I noticed an ordering issue - settings in this file were trying to be managed _before_ the PHP packages were being installed, and as such `/etc/php5` didn't exist (see #191)

This PR stops managing the file, but doesn't clean it up on systems that might already have it. I didn't feel that was safe.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -142,11 +142,13 @@ class php (
   anchor { 'php::end': }
 
   # Configure global PHP settings in php.ini
-  Anchor['php::begin'] ->
-  class {'::php::global':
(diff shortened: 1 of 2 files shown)


mmoll:
thanks @chrisboulton!

Issue #52, opened 2014-12-31
Unable to set extensions as "zend_extension"


parhamdoustdar:
Currently, it is not possible to change the `extension = ...` line that `puppet-php` inserts into the `extension-name.ini` file. When you want to install something like `xdebug`, for example, you need to change `extension = ...` to `zend_extension = ...`.

I currently override the contents of that file on every provision, but ideally, I'd not want to have to do that.

fpletz:
We should only do this if the extension was installed with pecl. If an extension from an operating system package is loaded the wrong way, please contact your package distributor to fix this.

I've implemented this in #54 which is to be released with 3.0.0 real soon now™.

parhamdoustdar:
Thank you a lot. No, I installed xdebug through pecl.

I'll be eagerly awaiting 3.0!

globin:
3.0 has since been released.

PR #193, opened 2016-06-09
Allow packaged extensions to be loaded as Zend extensions

chrisboulton:
# 54 (and #52) added support for Zend extensions, but as stated only allows them for PECL installed extensions. I don't think this is the right behavior, so this PR removes that restriction and allows any extension to be designated as a Zend extension.

Take for example Xdebug, which is packaged as `php5-xdebug` on Debian. Xdebug _WANTS_ to be loaded as a Zend extension, and won't work otherwise.

This PR is pretty opinionated in that it uses `php::config` to enforce either an extension or zend_extension value to be put into the mods-available configuration. Previously these lines were only managed for PECL extensions, or on FreeBSD. I don't see a reason to not do it for all extensions. This doesn't seem to cause any issues for me, but it'd be great to get some additional thought on the implications of this change.

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -39,7 +39,7 @@
 #
 # [*zend*]
 #  Boolean parameter, whether to load extension as zend_extension.
-#  This can only be set for pecl modules. Defaults to false.
+#  Defaults to false.
 #
 # [*settings*]
 #   Nested hash of global config parameters for php.ini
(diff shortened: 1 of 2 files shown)


jackdpeterson:
+1 to this PR. Need to load in xdebug in my local vagrant configs.

bastelfreak:
@chrisboulton thanks for the PR. Can you rebase please against latest master?

Vincent--:
Hi,

Could we have this PR merged please?
I can create another PR with the same content if needed

Regards,
Vincent

PR #195, opened 2016-06-11
Avoid puppets warning in puppetserver.log

jkroepke:
Hi,

puppet throws some warnings in the puppetserver.log (using 4.5.1), if puppet managed php an this host.

```
WARN  [qtp1416189715-65] [puppetserver] Puppet Arguments to Resource[] are all empty/undefined at /etc/puppetlabs/code/environments/productions/modules/php/manifests/extension.pp:99:13
```

I added a empty condition to define dependencies, if the header or compiler packages are required.

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -94,7 +94,7 @@ define php::extension(
     else {
       $real_package = "${package_prefix}${title}"
     }
-   
+
     unless empty($header_packages) {
       ensure_resource('package', $header_packages)
       Package[$header_packages] -> Package[$real_package]
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -94,10 +94,12 @@ define php::extension(
     else {
       $real_package = "${package_prefix}${title}"
     }
-
-    ensure_resource('package', $header_packages)
-    Package[$header_packages] -> Package[$real_package] -> ::Php::Config[$title]
-
+   
+    unless empty($header_packages) {
+      ensure_resource('package', $header_packages)
+      Package[$header_packages] -> Package[$real_package]
+    }
+    
     if $provider == 'pecl' {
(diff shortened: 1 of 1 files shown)


mmoll:
@jkroepke please use spaces instead of tab characters :mag:

manifests/extension.pp
```diff
@@ -95,9 +95,11 @@
       $real_package = "${package_prefix}${title}"
     }
 
-    ensure_resource('package', $header_packages)
-    Package[$header_packages] -> Package[$real_package] -> ::Php::Config[$title]
-
+    unless empty($header_packages) {
+      ensure_resource('package', $header_packages)
+      Package[$header_packages] -> Package[$real_package]
```
mmoll: I think `-> ::Php::Config[$title]` can stay here

manifests/extension.pp
```diff
@@ -180,8 +184,9 @@
 
   $config_root_ini = pick_default($::php::config_root_ini, $::php::params::config_root_ini)
   ::php::config { $title:
-    file   => "${config_root_ini}/${lowercase_title}.ini",
-    config => $final_settings,
+    file    => "${config_root_ini}/${lowercase_title}.ini",
+    config  => $final_settings,
+    require => Package[$real_package],
```
mmoll: I know it's working also here, but now having the dependency at all when this is empty is IMHO better.

manifests/extension.pp
jkroepke: if header_packages are empty, the dependency Package[$real_package] -> ::Php::Config[$title] is not defined.

mmoll: oh, this is about the headers, I see. thanks!

mmoll:
merged, danke @jkroepke!

Issue #184, opened 2016-05-03
Should we use LSB facts?


fpletz:
From https://github.com/mayflower/puppet-php/pull/180/files#r61703085

FutureSharks:
I think if it's definitely possible to avoid using these facts then why wouldn't we? It's just a needless dependency.

PR #198, opened 2016-06-12
stop using LSB facts

mmoll:
closes GH-184

The lsb usage in the tests can stay as-is, as we do get lsb facts always in the facterdb provided facts.

### DIFF
--- a/manifests/fpm.pp
+++ b/manifests/fpm.pp
@@ -95,8 +95,7 @@ class php::fpm (
 
   # Create an override to use a reload signal as trusty and utopic's
   # upstart version supports this
-  if $::osfamily == 'Debian' and
-    ($::lsbdistcodename == 'trusty' or $::lsbdistcodename == 'utopic') {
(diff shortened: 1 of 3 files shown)


manifests/fpm.pp
```diff
@@ -95,8 +95,7 @@
 
   # Create an override to use a reload signal as trusty and utopic's
   # upstart version supports this
-  if $::osfamily == 'Debian' and
-    ($::lsbdistcodename == 'trusty' or $::lsbdistcodename == 'utopic') {
+  if $::operatingsystem == 'Ubuntu' and ($::operatingsystemmajrelease == '14.04' or $::operatingsystemmajrelease == '14.10') {
```
igalic: how about
```puppet
versioncmp($::operatingsystemmajrelease, '14.04') >= 0)
```

mmoll: would that work with newer versions using systemd?

igalic: oh, right…
igore me, please

PR #200, opened 2016-06-14
Fixup naming of xml package for Ubuntu, when the package_prefix is

buzzdeee:
set as parameter to init.pp to some non-default

Also spec for xml package
add spec test for package_prefix

I'm in the phase of updating the php module, to get the
better php 7 support. Some of the boxes run on Ubuntu.
They have php 5.5 installed, and I don't want to upgrade
just because of the puppet php module update ;)

So besides other parameters, I set the package_prefix to 'php5-'
The only package that doesn't seem to catch that up is the php-xml
package, enforced via the php::dev class.

This PR also lets the xml package pick the right prefix, as well
as adding a bunch of spec tests, for ensuring correct functionality.

### DIFF
--- a/manifests/dev.pp
+++ b/manifests/dev.pp
@@ -27,7 +27,7 @@ class php::dev(
   }
 
   if $::operatingsystem == 'Ubuntu' {
-    ensure_packages(["${php::globals::package_prefix}xml"], {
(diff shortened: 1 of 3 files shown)


mmoll:
merged, danke @buzzdeee!
Special thanks for adding tests! :+1:

Issue #202, opened 2016-06-15
Specifying extensions can cause duplicate


dowlingw:
Specifying extensions can cause the manifest to fail to to apply to a host with the following error:

```
Jun 15 14:52:50 hostname puppet-agent[19003]: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Package[php7.0-xml] is already declared; cannot redeclare at /etc/puppetlabs/code/environments/production/modules/php/manifests/extension.pp:119 at /etc/puppetlabs/code/environments/production/modules/php/manifests/extension.pp:119:5 on node hostname
```

Using the following manifest:

```
        class { '::php':
                extensions => {
                        xml => {}
                }
        }
```

Suggest that the `php::extension` code migrate to use `ensure_packages` from `puppetlabs/puppetlabs-stdlib` as it looks like this is already used in places (and update documentation to note this depdendency)

mmoll:
just to double-check: is this on current master? Pull requests are always very elconme :)

dowlingw:
Whoops, sorry yes - Puppet 4 on Ubuntu 16.04 running mayflower/php master branch.

Having a poke around, it looks like the `php::pear` module is explicitly pulling in the xml module dependency for Ubuntu specifically:

```php
if $::operatingsystem == 'Ubuntu' {
    ensure_packages(["${php::package_prefix}xml"], {
      ensure  => present,
      require => Class['::apt::update'],
    })

    package { $package_name:
      ensure  => $ensure,
      require => [Class['::apt::update'],Class['::php::cli'],Package["${php::package_prefix}xml"]],
    }
  } else {
    package { $package_name:
      ensure  => $ensure,
      require => Class['::php::cli'],
    }
  }
```

Which isn't so bad, the use of `ensure_packages()` ensures that the include is safe. Except that this code runs first, so by the time the `php::extension` code runs, the package resource is already declared and it goes boom.

It's worth noting that the php-pear packages from this PPA seem to declare the dependancy in the first place, so it's probably unnecessary in _this_ case.

```bash
$ apt-cache depends php-pear
php-pear
  Depends: php-cli
    php5.5-cli
    php5.6-cli
    php7.0-cli
    php7.1-cli
  Depends: php-common
  Depends: php-xml
    php5.5-xml
    php5.6-xml
    php7.0-xml
    php7.1-xml
[...]

PR #203, opened 2016-06-16
#202 - Replace package resources with function call to ensure_packages

dowlingw:


mmoll:
merged, thanks @dowlingw!

PR #204, opened 2016-06-16
Allow to specify a fpm_service_provider,

buzzdeee:
to allow using non-default service provider for fpm.

I'm on ubuntu 14.04, where the fpm service provider
picks init as default, but it should use upstart.

The default for the service_provider are all set to 
undef, so it picks whatever is default, as it currently
is, only allows the user to override to whatever provider
he wants to use.


mmoll:
merged, danke @buzzdeee!

PR #206, opened 2016-06-18
remove superfluous requires

mmoll:
@mvisonneau @tdm4 could you have a look? From a quick glance, this should already be taken care of by the anchoring in init.pp.


tdm4:
@mmoll Looks fine to me. I've just tested it and it works. Go for it. 👍

mmoll:
@tdm4 thanks for testing!

mvisonneau:
@tdm4 lgtm, thanks a lot 👍

PR #224, opened 2016-07-03
fix error when dealing with php_version on suse

### LINKED ISSUE #223: SuSE Error: php::globals::php_version is not a hash or array when accessing it with 0
#220  introduces following error with puppet 3.8.x
```
 Error: php::globals::php_version is not a hash or array when accessing it with 0
```

xalopp:
closes  #223

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -58,7 +58,7 @@ class php::params inherits php::globals {
 
     'Suse': {
       if ($php::globals::php_version != undef) {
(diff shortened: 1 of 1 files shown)


xalopp:
@mmoll plz 👓 this PR.

Might require a test for the `$php_version_major` stuff.

mmoll:
merging this for unbreaking master... It's a shame we don't have more tests, but as always,

PR #225, opened 2016-07-15
fix trusty default php5-fpm / php5enmod name, disable xml dependency

marcofl:
this fixes fpm service name on Ubuntu Trusty 14.04 as there using default version.

### DIFF
--- a/manifests/dev.pp
+++ b/manifests/dev.pp
@@ -26,7 +26,7 @@ class php::dev(
     default   => $package,
   }
 
-  if $::operatingsystem == 'Ubuntu' {
+  if ($::operatingsystem == 'Ubuntu') and ($::php::globals::php_version != '5.5') {
     ensure_packages(["${php::package_prefix}xml"], {
       ensure  => present,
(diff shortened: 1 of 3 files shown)


mmoll:
@tdm4 could you have a look?

tdm4:
Only issue here, I think is, what if you install 5.5 from Ondřej Surý's repo instead of Ubuntu's base repos? Though I think, since we set Ubuntu systems to use that repo, this shouldn't be a problem.

mmoll:
OK, I'm getting the problem now... I'll reproduce as soon as I have time. If the PPA is enabled by default there should indeed be no problem. If the PPA is disabled, the other params need to get set to match, too.

mmoll:
Tested and as we do enable the PPA by default everything is fine.

If the PPA is explicitly disabled, the parameters in `::php::globals`  need to get explicitly overridden, too...

@marcofl could you open a PR to the README for that? Maybe around "Ubuntu systems and Ondřej's PPA"...

PR #226, opened 2016-07-19
Fix command to enable php extensions

iuriaranda:
Bug: phpquery command always exits with status code 0, even if the extension is not enabled, so the enable command is never run.
Fix: grep the string stating the extension is not enabled, and only run the enable command if that string is found

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -99,7 +99,7 @@ define php::extension(
       ensure_resource('package', $header_packages)
       Package[$header_packages] -> Package[$real_package]
     }
-    
+
     if $provider == 'pecl' {
       ensure_packages( [ $real_package ], {
         ensure   => $ensure,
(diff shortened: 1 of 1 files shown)


manifests/extension.pp
```diff
@@ -200,12 +200,12 @@
 
     if $sapi == 'ALL' {
       exec { $cmd:
-        unless  => "${ext_tool_query} -s cli -m ${lowercase_title}",
+        onlyif  => "${ext_tool_query} -s cli -m ${lowercase_title} | grep 'No module matches ${lowercase_title}'",
```
mmoll: please use `/bin/grep` here

manifests/extension.pp
```diff
@@ -200,12 +200,12 @@
...
         require =>::Php::Config[$title],
       }
     } else {
       exec { $cmd:
-        unless  => "${ext_tool_query} -s ${sapi} -m ${lowercase_title}",
+        onlyif  => "${ext_tool_query} -s ${sapi} -m ${lowercase_title} | grep 'No module matches ${lowercase_title}'",
```
mmoll: dito

mmoll:
Gràcies @iuriaranda!

iuriaranda:
De res @mmoll :wink:

mbrodala:
I just checked and this is actually not the case:

```
$ phpquery -v 5.6 -s cli -m xdebug
xdebug (Enabled for cli by local administrator)
$ echo $?
0
$ phpquery -v 5.6 -s cli -m xdebug2
No module matches xdebug2
$ echo $?
1
```

On which system was this checked? Using Ubuntu Trusty here.

iuriaranda:
In Xenial and PHP 7.0:

```
~# phpquery -v 7.0 -s cli -m zip
zip (Enabled for cli by maintainer script)
~# echo $?
0
~# phpquery -v 7.0 -s cli -m zip2
No module matches zip2
~# echo $?
0
```

Sorry for the late response btw.

PR #238, opened 2016-09-12
Rubocop fixes

alexjfisher:
Just a subset of the needed fixes here.


bastelfreak:
awesome commits are awesome.jpg

bastelfreak:
Hi @kmcfate,

we decided a few months back to drop ruby187 support. this is a really outdated ruby version that nobody should use anymore. The next release of the module will be a major release which drops ruby187 support.

bastelfreak:
Hi @kmcfate,
could you elaborate on the thumbs down emoji? Ruby187 is end of life since a very long time, it is known to be insecure and doesn't receive patches. Every modern distribution ships a newer ruby. Also Puppet(labs) offers bundled packages with ruby + openssl + puppet4. People in #voxpupuli and #puppet are happy to help everyone upgrade their legacy operatingsystem to something newer and/or upgrade puppet3/4.

daenney:
Considering we're soon moving to Puppet 4 only modules dropping Ruby 1.8.7 support is perfectly fine. It's not released yet and it'll be made clear that this change is coming. You shouldn't be running your production infrastructure of the master branch of this module so no one should get hurt by this being merged.

In general, I don't appreciate this just being down voted without so much as engaging in a conversation, especially when that happens by the original maintainers that have indicated no longer wishing to maintain the module in question.

mmoll:
As I already said in another discussion (I think it was in extlib), I do get the reasoning behind this, still this is making the module incompatible to setups, that are (unfortunately) still in wide use, like EL6 or SLES11 based OSes with Puppet 3. This is a general comment, not specific to puppet-php.

nibalizer:
When we talked at config mgmt camp 2016, we decided to allow branches to exist to keep old versions of puppet and ruby functional. If someone wants a r1.8 or p3.x branch, all they have to do is ask for it or create it.

mmoll:
@kmcfate :point_up:

PR #240, opened 2016-09-23
Rubocop fixes

alexjfisher:
Was 76 violations, now 19.


alexjfisher:
Make that 17.

alexjfisher:
15

alexjfisher:
10

lib/puppet/provider/package/pecl.rb
```diff
@@ -94,7 +94,7 @@ def install(useversion = true)
       end
     end
 
-    if pipe = @resource[:pipe]
+    if pipe == @resource[:pipe]
```
alexjfisher: @bastelfreak Oops.  This was clearly not right.

PR #241, opened 2016-09-24
Fix all remaining rubocop violations

alexjfisher:


alexjfisher:
Thanks to @elomatreb on IRC #ruby for the advice on https://github.com/voxpupuli/puppet-php/pull/241/commits/3f6cef97120dba393ff41a1d37bc4a196da3285a and https://github.com/voxpupuli/puppet-php/pull/241/commits/27af3722876d77a79437655b4dffd2787492454b

lib/puppet/provider/package/pear.rb
```diff
@@ -36,7 +36,7 @@ def self.pearlist(hash)
           pearhash[:provider] = :pear
           pearhash
         end
-      end.reject { |p| p.nil? }
+      end.reject(&:nil?)
```
alexjfisher: I was wondering if this could actually be replaced with `end.compact`, but was too scared!

bastelfreak:
LGTM but would be cool if somebody else could review that too.

lib/puppet/provider/package/pear.rb
elomatreb: `Array#compact` should work, the only difference to this manual solution I could find was that `compact` will only remove actual `nil`, whereas this will work on classes that define a custom `#nil?` (I see no reason why that would be something you want though).

lib/puppet/provider/package/pecl.rb
```diff
@@ -20,19 +20,17 @@ def self.pecllist(hash)
     begin
       list = execute(command).split("\n").map do |set|
         if hash[:justme]
-          if %r{^#{hash[:justme]}$}i.match(set)
+          if %r{^#{hash[:justme]}$}i =~ set
             if peclhash = peclsplit(set) # rubocop:disable Lint/AssignmentInCondition
```
elomatreb: Can these conditionals be merged into one if (to avoid nesting)?

alexjfisher: Good spot.  It certainly looks like it, doesn't it?  Lets look at the original code before any rubocop and refactoring started... https://github.com/voxpupuli/puppet-php/blob/4.0.0-beta1/lib/puppet/provider/package/pear.rb#L29

So assuming https://github.com/voxpupuli/puppet-php/pull/240/commits/6bbdc78177b53b84b6e6e7f27eb0edb21309e6ae was correct and `rubocop --auto-correct --only Style/EmptyElse` wasn't mistaken, then yes to merging the conditionals?

It's a real shame there are no spec tests covering this code...

elomatreb: The two versions should be functionally equivalent, as a non-branching if returns nil in Ruby.

I guess the real question is if this:

```ruby
if %r{^#{hash[:justme]}$}i =~ set
  if peclhash = peclsplit(set) # rubocop:disable Lint/AssignmentInCondition
    peclhash[:provider] = :peclcmd
    peclhash
  end
end
```
is more readable than:

```ruby
if %r{^#{hash[:justme]}$}i =~ set &&
   peclhash = peclsplit(set) # rubocop:disable Lint/AssignmentInCondition
  peclhash[:provider] = :peclcmd
  peclhash
end
```

alexjfisher:
I've added another commit with the extra refactoring.  Somebody might want to double check the diff of the two provider files from before I started touching them.
https://github.com/voxpupuli/puppet-php/compare/4.0.0-beta1...alexjfisher:even_more_rubocop?w=1#diff-cee9d5379fd01442b1ebb133bd077ccbL18
https://github.com/voxpupuli/puppet-php/compare/4.0.0-beta1...alexjfisher:even_more_rubocop?w=1#diff-cf9e2c386213badb7135250275e6ef43L17

PR #245, opened 2016-09-25
better gitignore

3flex:
Copies voxpupuli's gitignore from modulesync_config


alexjfisher:
@3flex Sure.  We might still be some way off doing a full modulesync.  This doesn't do any harm though.

PR #246, opened 2016-09-25
fix all strict_variables failures

### LINKED ISSUE #244: Module isn't 'strict variables' compatible
Currently, testing is done with `strict variables` turned off. https://github.com/voxpupuli/puppet-php/commit/5d50ac56414c4ef972a8aa6512032fb4c05891a0

3flex:
Closes #244 
- Instead of `php::fpm` pulling parameter values from `php`, they will be passed through from `php` to `php::fpm`
- Set `php::params::ext_tool_enable` and `php::params::ext_tool_query` to `undef` for all platforms other than Debian-derived. Previously these parameters only had a value set when running on Debian-derived distros. The only usage is in `php::extension`, which checks the module is running on Debian-derived distro anyway before using any `ext_tool_*` parameters.
- Changed two spec tests for defined types to `include ::php` as a pre-condition. This aligns with the module's README which heavily implies (though doesn't explicitly say) that to get started users should `include ::php`.


manifests/fpm.pp
```diff
@@ -44,12 +44,12 @@
 #   Defaults is empty hash.
 #
 class php::fpm (
-  $ensure               = $::php::ensure,
+  $ensure               = $::php::params::ensure,
```
alexjfisher: Not 100% sure about this.  It's certainly not backwards compatible.
Currently you could do
```puppet
class {'php': ensure => absent}
```

I started looking at his the other day.  I think the problem is the way the class is tested.  It's supposed to be a private class only called from `init.pp`.  In the testing, it's tested as if it was independent.  You've probably noticed you can't just do a
```ruby
let(:pre_condition) { 'include ::php' }
```
without getting a duplicate resource declaration.  I thought about doing
```ruby
let(:pre_condition) { 'class { "::php": fpm => false' }
```
but that seems pretty hacky too.  Maybe the tests should just be moved into `php_spec.rb`?

alexjfisher: @3flex Oops.  I see you've covered that later on.

manifests/init.pp
```diff
@@ -159,13 +159,15 @@
   if $fpm {
     Anchor['php::begin'] ->
       class { '::php::fpm':
+        ensure           => $ensure,
```
alexjfisher: if `fpm.pp` is only ever declared from `init.pp` should it even have defaults for it's parameters?  The defaults never get used, do they?

manifests/fpm.pp
```diff
@@ -44,12 +44,12 @@
...
   $service_ensure       = $::php::params::fpm_service_ensure,
   $service_enable       = $::php::params::fpm_service_enable,
   $service_name         = $::php::params::fpm_service_name,
   $service_provider     = undef,
```
alexjfisher: It'd be nice if there was some consistency in this class.  Why does `service_provider` default to `undef` whilst `service_name` defaults to what's in `params.pp`?

manifests/init.pp
alexjfisher: @3flex I've opened https://github.com/voxpupuli/puppet-php/pull/248 with a possible way we could go about refactoring.  Let me know what you think.

3flex: Doesn't look like it... would there be any other reason to leave them in there?

There are 5 php::fpm parameters that aren't set by ::php currently. Do you think those should be stripped out, then the remaining 7 default to undef, and pass through the values from ::php?

3flex: Whoops didn't see your comment or #248 until after I commented. Looks cleaner than what I've proposed.

bastelfreak:
Hi @3flex, thanks for the great PR! Can you please rebase?

3flex:
@bastelfreak I went ahead and rebased. Not sure if you want to merge or want something more complete like #248 instead.

bastelfreak:
Very good question. We should discuss it with @alexjfisher.

PR #247, opened 2016-09-25
fixtures: don't pin puppetlabs-apt version

3flex:
Previously this was frozen to 1.8.x, released about 18 months ago.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,9 +1,7 @@
 fixtures:
   repositories:
     stdlib: "git://github.com/puppetlabs/puppetlabs-stdlib.git"
-    apt:
-      repo: "git://github.com/puppetlabs/puppetlabs-apt.git"
-      branch: "1.8.x"
+    apt: "git://github.com/puppetlabs/puppetlabs-apt.git"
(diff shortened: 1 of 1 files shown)


alexjfisher:
Not sure about this. I think most other modules get fixtures from git.
eg
https://github.com/voxpupuli/puppet-archive/blob/master/.fixtures.yml
https://github.com/voxpupuli/puppet-corosync/blob/master/.fixtures.yml
https://github.com/voxpupuli/puppet-jira/blob/master/.fixtures.yml
https://github.com/voxpupuli/puppet-mcollective/blob/master/.fixtures.yml
https://github.com/voxpupuli/puppet-collectd/blob/master/.fixtures.yml

Those are the first 5 I picked at semi-random (they were in my history)

alexjfisher:
@3flex I think you're right about the `apt` module though.

The metadata has the following.
```
{ "name": "puppetlabs/apt", "version_requirement": ">= 1.8.0 < 3.0.0" }
```

alexjfisher:
@3flex Would you like to rename the PR before we merge?

3flex:
Done!

Issue #151, opened 2015-12-21
Remove "private class" checks


tobru:
In my personal opinion it doesn't make sense that classes check if the were called from the same module. I'm speaking of checks like this: `if $caller_module_name != $module_name`. I'd like to remove this checks to make the module more flexible. If there are no strong reasons to leave this there, I'm happy to create a PR.

mpdude:
I think it is important to make a difference between "public" classes and "private" ones. Puppet does not support a better way to enforce encapulation here.

For the maintainers, it might be important to know which parameters may be changed (renamed, moved to other internal classes) and whihc must be kept as part of the module API.

tobru:
Yes of course, but if I want to use a "private" class directly, f.e. because the main class does not allow to configure all parameters of a private class, or I have other ideas to use a private class directly, then it's annoying when there are a lot of red warnings. If I use a private class directly, then I'm also aware of the fact that parameters could change which then is my own responsibility to fix my class. Maybe put a notice in the class documentation which says that this class could change parameters from time to time...? And mention it in the changelog if they change?

The only way of making the usage of a private class not needed is when all parameters would be exposed to the main class, but that misses a bit the point of a private class.

It would make this module much more user friendly without this private class enforcement, I could call the classes from outside like it fits my needs. And I think flexibility is very important for a "base" module like this one.

fpletz:
We wrote this module with hiera usage in mind. The idea is to only include the main class and configure the rest with hiera. Clearly, seeing many issues on that topic, people prefer to mix data and code and I agree that we shouldn't stand in the way here.

Not that there is some imported logic and dependency management happening in `init.pp` that you would have to replicate currently. That's why most sub-classes are marked as private. Fixing this is probaly a major refactoring which we are not able to do right now due to time constraints.

Pull requests are welcome.

PR #248, opened 2016-09-25
RFC Refactor fpm class

alexjfisher:
DO NOT MERGE. Just for discussion.

### DIFF
--- a/manifests/fpm.pp
+++ b/manifests/fpm.pp
@@ -45,20 +45,20 @@
 #
 class php::fpm (
   $ensure               = $::php::ensure,
-  $service_ensure       = $::php::params::fpm_service_ensure,
-  $service_enable       = $::php::params::fpm_service_enable,
-  $service_name         = $::php::params::fpm_service_name,
-  $service_provider     = undef,
-  $package              = "${::php::package_prefix}${::php::params::fpm_package_suffix}",
-  $inifile              = $::php::params::fpm_inifile,
-  $settings             = {},
-  $global_pool_settings = {},
-  $pools                = { 'www' => {} },
-  $log_owner            = $::php::params::fpm_user,
-  $log_group            = $::php::params::fpm_group,
-) inherits ::php::params {
+  $service_ensure       = $::php::fpm_service_ensure,
+  $service_enable       = $::php::fpm_service_enable,
+  $service_name         = $::php::fpm_service_name,
+  $service_provider     = $::php::fpm_service_provider,
+  $package              = $::php::real_fpm_package,
+  $inifile              = $::php::fpm_inifile,
+  $settings             = $::php::real_settings,
+  $global_pool_settings = $::php::real_fpm_global_pool_settings,
+  $pools                = $::php::real_fpm_pools,
+  $log_owner            = $::php::log_owner,
+  $log_group            = $::php::log_group,
+) {
 
-  if $caller_module_name != $module_name {
+  if ! defined(Class['php']) {
     warning('php::fpm is private')
   }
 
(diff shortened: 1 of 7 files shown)


alexjfisher:
Incomplete, untested, would need tests refactored etc.

manifests/fpm.pp
```diff
@@ -45,21 +45,21 @@
 #
 class php::fpm (
   $ensure               = $::php::ensure,
-  $service_ensure       = $::php::params::fpm_service_ensure,
-  $service_enable       = $::php::params::fpm_service_enable,
-  $service_name         = $::php::params::fpm_service_name,
-  $service_provider     = undef,
```
alexjfisher: some parameters defaulted to `params.pp`, others to `undef` or `{}`. If it's a private class only being declared from `::php` this didn't make much sense.

manifests/fpm.pp
```diff
@@ -45,21 +45,21 @@
...
+  $service_provider     = $::php::fpm_service_provider,
   $package              = "${::php::package_prefix}${::php::params::fpm_package_suffix}",
   $inifile              = $::php::params::fpm_inifile,
-  $settings             = {},
+  $settings             = $::php::real_settings,
   $global_pool_settings = {},
   $pools                = { 'www' => {} },
-  $log_owner            = $::php::params::fpm_user,
-  $log_group            = $::php::params::fpm_group,
-) inherits ::php::params {
```
alexjfisher: no need to inherit params.pp *if* we can only get declared from the base class.

manifests/fpm.pp
```diff
@@ -45,21 +45,21 @@
...
+  $log_owner            = $::php::log_owner,
+  $log_group            = $::php::log_group,
+) {
 
   if $caller_module_name != $module_name {
-    warning('php::fpm is private')
+    fail('php::fpm is private')
```
alexjfisher: What good is a warning?  If it really is private, we should `fail` instead.

manifests/fpm/pool.pp
```diff
@@ -156,7 +156,10 @@
   $base_dir                  = undef,
 ) {
 
-  include ::php::params
```
alexjfisher: It's a bit late to `include ::php::params` here.  If it's not already been included, the type default for `root_group` would be missing.

manifests/fpm/pool.pp
```diff
@@ -156,7 +156,10 @@
...
+  # The base class must be included first because it is used by parameter defaults
```
alexjfisher: Shamelessly stolen from https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/vhost.pp

manifests/init.pp
```diff
@@ -156,19 +156,7 @@
...
-        service_ensure   => $fpm_service_ensure,
-        service_name     => $fpm_service_name,
-        service_provider => $fpm_service_provider,
-        settings         => $real_settings,
-        log_owner        => $log_owner,
-        log_group        => $log_group,
-      } ->
-    Anchor['php::end']
-  }
+  if $fpm { contain '::php::fpm' }
```
alexjfisher: All these parameters will now just default correctly to the values from the base class.  You can now override them in hiera if you really want.  We're probably alright switching to `contain` now (puppet >= 3.4)?

manifests/fpm.pp
3flex: There was an issue opened to remove these checks: https://github.com/voxpupuli/puppet-php/issues/151, so changing these to fail will definitely upset some people.

alexjfisher: I guess all of the parameters should be exposed in the base class, otherwise you *have* to use hiera to set them.  What about switching to an `if ! defined(Class['php'])` here too?  If you really wanted to declare `php::fpm` yourself, you still could as long as you declared `php` first with `fpm => false`.

It's going to be a new major release, and the class was marked as private, so people can't complain too much if we make a backwards incompatible change?  After all, the class was actively warning them not to declare it themselves!

igalic: how long do you reckon will this rework take?
we could have a lot of bug-fix releases in the meantime…

alexjfisher: @igalic PR updated with the suggested rework.  `php::fpm::config` could do with a bunch of refactoring too, but that's a bigger task.

manifests/fpm.pp
```diff
@@ -45,20 +45,20 @@
...
+  $settings             = $::php::real_settings,
+  $global_pool_settings = $::php::real_fpm_global_pool_settings,
+  $pools                = $::php::real_fpm_pools,
+  $log_owner            = $::php::log_owner,
+  $log_group            = $::php::log_group,
+) {
 
-  if $caller_module_name != $module_name {
+  if ! defined(Class['php']) {
     warning('php::fpm is private')
```
Vincent--: Not sure the warning message is still matching the test. If you include the ```php``` class in your manifest, you can now use the ```php::fpm``` class which is then no more ```private``` (but I might be wrong)

PR #249, opened 2016-09-29
Fix dependency for extentions with no package source

bisscuitt:
This fixes an issue where setting `source => 'none'` to a `php::extension` resource results in the following error:
```
Error: Could not find dependency Package[undef] for Php::Config[json] at /etc/puppet/modules/php/manifests/extension.pp:190
```

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -88,6 +88,7 @@ define php::extension(
   }
 
   if $provider != 'none' {
+
     if $provider == 'pecl' {
       $real_package = "pecl-${title}"
     }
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @bisscuitt, thanks for the PR. This will fail with strict vars because $package_depends is not always defined. Please define the fallback to undef. Also can you please add tests?

bisscuitt:
I've updated the code so that ` $package_depends` now falls back to `undef`.

There is now an additional test, but I'm pretty new to testing in puppet so not 100% sure it's what you are loooking for. Any guidance would be greatly appreciated.

jonhattan:
The error on my side was:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: Php::Config[pdo] { require => Package[undef] }, because Package[undef] doesn't seem to be in the catalog
```

Confirming this MR fixes it. Thanks!

bastelfreak:
Thanks!

PR #250, opened 2016-09-29
Fixapt

cliff-wakefield:
Under a Debian like environment apt update is called but apt is **not** included.


bastelfreak:
Hi @cliff-wakefield, thanks for the PR!

PR #251, opened 2016-09-29
Update module Ubuntu 14.04 default to official repository setup

jk2l:
Fix for Ubuntu 14.04 Php installation:

Summary:
* The currents rule only work for PPA. Using 14.04 Official Ubuntu repository does not match any of the rules. 
* Removed 14.10 from metadata. It is in EOL and thing can randomly break. 
* Removed 12.04 from metadata. Default installation is 5.3, this module won't support any of it

### DIFF
--- a/manifests/dev.pp
+++ b/manifests/dev.pp
@@ -26,7 +26,8 @@ class php::dev(
     default   => $package,
   }
 
-  if $::operatingsystem == 'Ubuntu' {
+  # Default PHP come with xml module and no seperate package for it
+  if $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease >= '16.04'  {
     ensure_packages(["${php::package_prefix}xml"], {
       ensure  => present,
       require => Class['::apt::update'],
--- a/manifests/dev.pp
+++ b/manifests/dev.pp
@@ -27,7 +27,7 @@ class php::dev(
   }
 
   # Default PHP come with xml module and no seperate package for it
-  if $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease >= '16.04'  {
+  if $::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '16.04') >= 0  {
     ensure_packages(["${php::package_prefix}xml"], {
       ensure  => present,
       require => Class['::apt::update'],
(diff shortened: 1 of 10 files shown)


jk2l:
The module tested against both 14.04 and 16.04 ubuntu. But our test case will need to patch according to it as almost everywhere hardcoded for php5-

manifests/dev.pp
```diff
@@ -26,7 +26,8 @@
     default   => $package,
   }
 
-  if $::operatingsystem == 'Ubuntu' {
+  # Default PHP come with xml module and no seperate package for it
+  if $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease >= '16.04'  {
```
bastelfreak: can you use versioncompare here instead of >=?

manifests/pear.pp
```diff
@@ -43,7 +43,8 @@
   validate_string($ensure)
   validate_string($package_name)
 
-  if $::operatingsystem == 'Ubuntu' {
+  # Default PHP come with xml module and no seperate package for it
+  if $::operatingsystem == 'Ubuntu' and $::operatingsystemrelease >= '16.04' {
```
bastelfreak: also versioncompare here

spec/classes/php_fpm_spec.rb
```diff
@@ -9,16 +9,9 @@
...
-          when 'Ubuntu'
-            let(:params) { { package: 'php5.6-fpm', ensure: 'latest' } }
-            it { is_expected.to contain_package('php5.6-fpm').with_ensure('latest') }
-            it { is_expected.to contain_service('php5.6-fpm').with_ensure('running') }
-          when 'Debian'
-            let(:params) { { package: 'php5-fpm', ensure: 'latest' } }
-            it { is_expected.to contain_package('php5-fpm').with_ensure('latest') }
-            it { is_expected.to contain_service('php5-fpm').with_ensure('running') }
-          end
+          let(:params) { { package: 'php5-fpm', ensure: 'latest' } }
```
bastelfreak: can you migrate the let block to multiline do/end?

spec/defines/extension_spec.rb
```diff
@@ -6,17 +6,15 @@
       let :facts do
         facts
       end
-      let(:pre_condition) { 'include php::params' }
+      let(:pre_condition) { '
```
bastelfreak: replace {} with do/end

bastelfreak:
versioncmp() is documented here https://docs.puppet.com/puppet/latest/reference/function.html#versioncmp

PR #252, opened 2016-10-04
fix php5enmod / php5query for Ubuntu Trusty and php5.5

marcofl:
on trusty with base php 5.5, the php5enmod and php5query commands are:
'/usr/sbin/php5enmod'
'/usr/sbin/php5query'

so I think it's the best to use the same case as for php 5.4 - not sure if php 5.4 is even supported / used though any longer on Trusty.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -44,7 +44,7 @@ class php::globals (
     'Debian': {
       if $::operatingsystem == 'Ubuntu' {
         case $globals_php_version {
-          /^5\.4/: {
+          /^5\.[45]/: {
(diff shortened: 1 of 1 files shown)


jk2l:
this issue should be fixed in https://github.com/voxpupuli/puppet-php/pull/251

right now using manage_repo = false, should point it to use 5.x which end up into "default" section

jk2l:
this issue should be fixed in #251

right now using manage_repo = false, should point it to use 5.x which end up into "default" section

juniorsysadmin:
Fixed by #251 , closing

PR #258, opened 2016-10-20
Allow packaged extensions to be loaded as Zend extensions (bis)

Vincent--:
Thanks to [chrisboulton](https://github.com/chrisboulton), see https://github.com/voxpupuli/puppet-php/pull/193

https://github.com/voxpupuli/puppet-php/pull/54 (and https://github.com/voxpupuli/puppet-php/issues/52) added support for Zend extensions, but as stated only allows them for PECL installed extensions. I don't think this is the right behavior, so this PR removes that restriction and allows any extension to be designated as a Zend extension.

Take for example Xdebug, which is packaged as php5-xdebug on Debian. Xdebug WANTS to be loaded as a Zend extension, and won't work otherwise.

This PR is pretty opinionated in that it uses php::config to enforce either an extension or zend_extension value to be put into the mods-available configuration. Previously these lines were only managed for PECL extensions, or on FreeBSD. I don't see a reason to not do it for all extensions. This doesn't seem to cause any issues for me, but it'd be great to get some additional thought on the implications of this change.

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -39,7 +39,7 @@
 #
 # [*zend*]
 #  Boolean parameter, whether to load extension as zend_extension.
-#  This can only be set for pecl modules. Defaults to false.
+#  Defaults to false.
 #
 # [*settings*]
 #   Nested hash of global config parameters for php.ini
@@ -130,10 +130,6 @@ define php::extension(
     $package_depends = undef
   }
 
-  if $provider != 'pecl' and $zend {
-    fail('You can only use the zend parameter for pecl PHP extensions!')
-  }
-
   if $zend == true {
     $extension_key = 'zend_extension'
     if $php_api_version != undef {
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @Vincent--, thanks for the PR, can you please take a look at the failing spec tests?

Vincent--:
I cannot see the result on Travis, I got this error message: "The repository at voxpupuli/puppet-php was not found." Is it only me or a global issue?

dhoppe:
@Vincent-- This is because of the DNS outage at the US east coast. I will restart the TravisCI build.

Vincent--:
Thank you, @dhoppe 👍 

I've corrected the issue, but happy if someone can double-check as it was my first time with rspec...
Also unrelated newbie question: why ```bundle exec rake``` fails with a lot of error on my local machine but runs successfully on Travis?

Regards,
Vincent.

Vincent--:
Any update on that? :-)
We really need this in the mainstream 👍

rick-cti:
This introduced a problem in our scenario - Ubuntu 14.04, PHP 5.6 and the "xdebug" extension installed.

Before this commit, the PHP apt package correctly configured the `xdebug.ini` file:

`zend_extension = xdebug.so`

However, since this commit, the Puppet module **also** tries to manage the file, adding in an extra line:

`zend_extension = xdebug.so
extension = xdebug.so`

This second line causes PHP requests to loop endlessly until the Linux OOM killer ends the process.

The fix here was to set `zend = true` as a parameter to the `php::extension` class for the `xdebug` extension.

I also believe this affects the {{mysql}} extension on PHP 7.0; by default, the apt package doesn't enable this extension for PHP 7.0 (as it's no longer used by version 7.0+), but the Puppet module now enables it regardless.

PR #265, opened 2016-11-03
Ability to manage my own curl

edestecd:
We want to manage curl with different params like ensure => latest...
So ensure_packages no work for us.


dhoppe:
Thank you very much for your contribution. I will merge this PR, because it does not change how the module works, but adds more flexibility for the users.

igalic:
→ #273 & #274.

Issue #273, opened 2016-11-28
php::composer should reuse archive, rather than managing curl


igalic:
### Affected Puppet, Ruby, OS and module versions/distributions

- Module version: 4.0.0-betaX

### How to reproduce (e.g Puppet code you use)

```
include php::composer
```

### What are you seeing

```
Error: Evaluation Error: Error while evaluating a Function Call, Duplicate declaration: 
  Package[curl] is already declared in file
  /etc/puppetlabs/code/environments/production/site/profiles/manifests/extra_software.pp:4;
  cannot redeclare at /etc/puppetlabs/code/environments/production/modules/php/manifests/composer.pp:45:21 on node app02.esat
```
### What behaviour did you expect instead

I would *not* expect a module named `php` to manage the installation of cURL. I'd expect our module to reuse @voxpupuli - [archive](https://github.com/voxpupuli/puppet-archive)

igalic:
the fact that this particular class also sets (or rather, unsets) an `environment` variable makes it almost impossible to override the settings in hiera…

igalic:
https://github.com/voxpupuli/puppet-php/pull/264 ← this pr brought forth the above behaviour, re `environment`

edestecd:
I wholeheartedly agree! Code reuse is better.

juniorsysadmin:
My understanding is that this is fixed by #274 
If the change has introduced a bug, that's another matter

PR #274, opened 2016-11-29
use voxpupuli/archive to download composer

igalic:
rather than using a hand-crafted curl call, we use voxpupuli/archive to
download composer. This has the benefit that it doesn't require us to
$manage_curl. But we also remove `$environment`! This variable name is
poorly chosen. Overriding it (or even just setting it to `undef`), could
mean that someone's `hiera()` lookups are failing, as soon as they enter
our module!

This pull-request addresses #273.

### DIFF
--- a/manifests/composer.pp
+++ b/manifests/composer.pp
@@ -8,11 +8,11 @@
 # [*path*]
 #   Holds path to the Composer executable
 #
-# [*environment*]
-#   Environment variables for settings such as http_proxy, https_proxy, or ftp_proxy
+# [*proxy_type*]
+#    proxy server type (none|http|https|ftp)
 #
-# [*manage_curl*]
-#   Should we ensure curl is installed or do you want to manage that?
+# [*proxy_server*]
+#   specify a proxy server, with port number if needed. ie: https://example.com:8080.
 #
 # [*auto_update*]
 #   Defines if composer should be auto updated
@@ -24,13 +24,13 @@
 #   UNIX group of the root user
 #
 class php::composer (
-  $source      = $::php::params::composer_source,
-  $path        = $::php::params::composer_path,
-  $environment = undef,
-  $manage_curl = true,
-  $auto_update = true,
-  $max_age     = $::php::params::composer_max_age,
-  $root_group  = $::php::params::root_group,
+  $source       = $::php::params::composer_source,
+  $path         = $::php::params::composer_path,
+  $proxy_type   = undef,
+  $proxy_server = undef,
+  $auto_update  = true,
+  $max_age      = $::php::params::composer_max_age,
+  $root_group   = $::php::params::root_group,
 ) inherits ::php::params {
 
   if $caller_module_name != $module_name {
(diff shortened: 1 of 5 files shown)


manifests/composer.pp
```diff
@@ -42,15 +42,12 @@
...
-    creates     => $path,
-    path        => ['/bin/', '/sbin/' , '/usr/bin/', '/usr/sbin/',
-                    '/usr/local/bin', '/usr/local/sbin'],
-    require     => [Class['::php::cli'],Package['curl']],
+  archive { 'download composer':
+    source       => $source,
+    creates      => $path,
+    proxy_type   => $proxy_type,
+    proxy_server => $proxy_server,
+    require      => [Class['::php::cli'],Package['curl']],
```
alexjfisher: require Package['curl'] not needed now

manifests/composer/auto_update.pp
```diff
@@ -25,16 +29,22 @@
...
+  $proxy_type   = undef,
+  $proxy_server = undef,
 ) {
 
   if $caller_module_name != $module_name {
     warning('php::composer::auto_update is private')
   }
 
+  $env = $proxy_type? {
+    undef   => {},
```
alexjfisher: It appears this isn't puppet 3 compatible. From https://docs.puppet.com/puppet/3.8/reference/lang_conditional.html#values

> Values may be any of the following:

> Any literal value, with the exception of hash literals

igalic: ow

igalic: i'll have to return `undef` then.

manifests/composer.pp
```diff
@@ -42,15 +42,12 @@
...
-    creates     => $path,
-    path        => ['/bin/', '/sbin/' , '/usr/bin/', '/usr/sbin/',
-                    '/usr/local/bin', '/usr/local/sbin'],
-    require     => [Class['::php::cli'],Package['curl']],
+  archive { 'download composer':
+    source       => $source,
+    creates      => $path,
+    proxy_type   => $proxy_type,
+    proxy_server => $proxy_server,
+    require      => Class['::php::cli'],
```
alexjfisher: This dependency looks a bit odd.  But it was there before, so maybe removing it will have unintended consequences.

igalic: composer requires php to run, so this makes sense.

alexjfisher: to run, but not to download.  Maybe it should just be `include`d instead?

edestecd:
$environment is not needed with archive, because archive has specific proxy settings. Just make sure you pass them through so we can set them, when you make the change.

igalic:
@edestecd 
that's why i removed `environment`, however, as i hint at in https://github.com/voxpupuli/puppet-archive/issues/244, there's room for improval in archive ;)

igalic:
```
Error: Parameter path failed on Archive[download composer]: archive path must be absolute: download composer at /etc/puppetlabs/code/environments/production/modules/php/manifests/composer.pp:45
```

PR #278, opened 2016-12-06
respect $manage_repos, do not include ::apt if set to false

fraenki:
PR #250 introduced code that unconditionally includes `::apt`. This should be changed to respect wether `$manage_repos` is actually set to `true` or not.

Otherwise I'm unable to use my own repo management, because the include leads to a duplicate declaration:

```
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Class[Apt] is already declared
```

### DIFF
--- a/manifests/packages.pp
+++ b/manifests/packages.pp
@@ -14,6 +14,7 @@
 #
 class php::packages (
   $ensure          = $::php::ensure,
+  $manage_repos    = $::php::manage_repos,
   $names_to_prefix = prefix(
     $::php::params::common_package_suffixes, $::php::package_prefix # lint:ignore:parameter_documentation
   ),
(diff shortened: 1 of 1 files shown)


dhoppe:
@fraenki Thank you very much for your contribution.

joshuaspence:
Sorry, I commented on the commit without seeing the corresponding pull request. I'll fix a PR.

roll4life:
@joshuaspence tested your fix on ubuntu 16.04, and it does resolve the issue with php7.0-cli, but if you have `dev => true` or `pear => true` it throws `Error: Could not find dependency Class[Apt::Update] for Package[php7.0-xml]`

joshuaspence:
When you say that you tested my fix, did you mean #323?

roll4life:
Yes, that is correct.  Sorry I should have specified.

roll4life:
correct.  That is an issue of `include ::apt` missing from manifests/dev.pp and manifests/pear.pp
at least in puppet 3.  I can create a separate issue if needed.

PR #281, opened 2016-12-21
modulesync 0.16.4

bastelfreak:


alexjfisher:
Coverage drop is expected because we're now doing `track_files 'lib/**/*.rb'`

PR #293, opened 2017-01-16
Add PHP 7.1 support on Debian

fstr:
I've added support for PHP 7.1 on Debian (jessie, stretch).

```
PHP 7.1.0-5+0~20161222133327.13+jessie~1.gbp46a191 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.1.0-dev, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.1.0-5+0~20161222133327.13+jessie~1.gbp46a191, Copyright (c) 1999-2016, by Zend Technologies
    with Xdebug v2.5.0, Copyright (c) 2002-2016, by Derick Rethans
```

I'm not sure if it's the most elegant way to install the required packages and how to avoid name conflicts.

Feedback is welcome.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -66,7 +66,7 @@ class php::globals (
         }
       } else {
         case $globals_php_version {
-          /^7/: {
+          /^7\.[0-9]/: {
             $default_config_root  = "/etc/php/${globals_php_version}"
             $default_fpm_pid_file = "/var/run/php/php${globals_php_version}-fpm.pid"
             $fpm_error_log        = "/var/log/php${globals_php_version}-fpm.log"
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -69,4 +69,26 @@ class php::repo::debian(
       }
     }
   }
+
+  if ($php::globals::php_version == '7.1') {
+    # Required packages for PHP 7.1 repository
+    ensure_packages(['lsb-release', 'ca-certificates'], {'ensure' => 'present'})
+
+    # Add PHP 7.1 key + repository
+    apt::key { 'php::repo::debian-php71':
+      key        => 'DF3D585DB8F0EB658690A554AC0E47584A7A714D',
+      key_source => 'https://packages.sury.org/php/apt.gpg',
+    }
+
+    ::apt::source { 'source_php_71':
+      location    => 'https://packages.sury.org/php/',
+      release     => $::lsbdistcodename,
+      repos       => 'main',
+      include_src => false,
+      require     => [
+        Apt::Key['php::repo::debian-php71'],
+        Package['apt-transport-https', 'lsb-release', 'ca-certificates']
+      ],
+    }
+  }
 }
--- a/metadata.json
+++ b/metadata.json
@@ -15,7 +15,7 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": ">= 4.1.0 < 5.0.0"
+      "version_requirement": ">= 4.4.0 < 5.0.0"
     },
     {
       "name": "puppetlabs/inifile",
--- a/metadata.json
+++ b/metadata.json
@@ -11,7 +11,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 5.0.0"
+      "version_requirement": ">= 4.16.0 < 5.0.0"
     },
     {
       "name": "puppetlabs/apt",
--- a/spec/classes/php_repo_spec.rb
+++ b/spec/classes/php_repo_spec.rb
@@ -7,6 +7,10 @@ describe 'php::repo', type: :class do
         facts
       end
 
+      let :pre_condition do
+        'include php'
+      end
+
       describe 'when configuring a package repo' do
         case facts[:osfamily]
         when 'Debian'


manifests/repo/debian.pp
```diff
@@ -62,4 +62,35 @@
       }
     }
   }
+
+  if ($php::globals::php_version == '7.1') {
+    # Required packages for PHP 7.1 repository
+    package { 'apt-transport-https':
```
vinzent: I do see that this package is required, but this could break others configuration if they defined the package in a profile. 

the same for the other packages declared here.

I really would not do business of adding repos. and such packags. If doing the roles/profiles pattern it should be solved in the profile level. For me this module should not try to add repos. but thats my personal opinion. so many things fail if you would run such things in a restricted environment (no internet access, and the policy to use internal repos).

Right now I can't say what to do with these packages. :-/

manifests/repo/debian.pp
```diff
@@ -62,4 +62,35 @@
...
+    package { 'lsb-release':
+      ensure => present,
+    }
+
+    package { 'ca-certificates':
+      ensure => present,
+    }
+
+    # Add PHP 7.1 key + repository
+    create_resources(::apt::key, { 'php::repo::debian-php71' => {
```
vinzent: why using `create_resources()` here? why not just plain puppet code? 

```puppet
apt::key { 'php::repo::debian-php71':
  key => 'DF3D585DB8F0EB658690A554AC0E47584A7A714D', 
  key_source => 'https://packages.sury.org/php/apt.gpg',
}
```

spec/spec_helper.rb
```diff
@@ -24,8 +24,8 @@
     puppetversion: Puppet.version,
     facterversion: Facter.version
   }
-  default_facts.merge!(YAML.load(File.read(File.expand_path('../default_facts.yml', __FILE__)))) if File.exist?(File.expand_path('../default_facts.yml', __FILE__))
-  default_facts.merge!(YAML.load(File.read(File.expand_path('../default_module_facts.yml', __FILE__)))) if File.exist?(File.expand_path('../default_module_facts.yml', __FILE__))
+  default_facts.merge!(YAML.safe_load(File.read(File.expand_path('../default_facts.yml', __FILE__)))) if File.exist?(File.expand_path('../default_facts.yml', __FILE__))
```
vinzent: the safe_load rubocop was decided to be blacklisted. the load should be used.

manifests/globals.pp
```diff
@@ -76,14 +76,14 @@
         }
       } else {
         case $globals_php_version {
-          /^7/: {
+          /^7\.[0-9]/: {
```
vinzent: what problem is solved by changing the regex?

manifests/repo/debian.pp
fstr: True. Just copied from above. Can be plain puppet code.

manifests/repo/debian.pp
fstr: I see conflict potential as well in the packages. So I'm also not sure.

If adding repos is not part of this module, then what's the purpose of this file `debian.pp`?

manifests/globals.pp
fstr: Being able to use:
`$package_prefix       = "php${globals_php_version}-"`

Because the package prefix is `php7.0-` and not `php7-`
Before that change it was hardcoded.

manifests/repo/debian.pp
vinzent: I'm not the original author of the module. I think it solves problems he or the users have. ;-)

oranenj: Many modules solve this by having a "manage_repo" parameter in the init class.

You should probably also use ensure_packages to avoid conflicts with existing profiles.

oranenj: never mind, I noticed that this is indeed gated by such a boolean. Should be fine then.

smuggli:
For all those who were waiting for a dotdeb release.

https://www.dotdeb.org/2017/01/27/php-7-1-dotdeb/

.rubocop.yml
```diff
@@ -302,7 +302,7 @@ Style/EmptyLiteral:
 Metrics/LineLength:
   Enabled: False
 
-Style/MethodCallParentheses:
+Style/MethodCallWithoutArgsParentheses:
```
jyaworski: Will modulesync overwrite this? @bastelfreak

vinzent: @jyaworski  no, its already in modulesync_config - https://github.com/voxpupuli/modulesync_config/blob/master/moduleroot/.rubocop.yml#L310

manifests/repo/debian.pp
sts: Please use [stdlib::ensure_packages](https://github.com/puppetlabs/puppetlabs-stdlib#ensure_packages) for this.

ffrank:
Hi @fstr is this still on your plate? Do you need anything from our end?

jonhattan:
This needs a reroll. The fix for package_prefix was already committed in https://github.com/voxpupuli/puppet-php/pull/329

bastelfreak:
I rebased the branch and implemented the changes @Vincent-- mentioned.

jonhattan:
Sorry I don't think this PR is the proper solution. At the end we want dotdeb for wheezy (php 5.6) and sury for jessie and stretch (5.6, 7.0, 7.1 and 7.2 as of now). I'll try to propose a PR in the coming days.

c33s:
@jonhattan you are right, we want dotdeb for wheezy and sury for jessie and strech. ping me if you start working on a PR

this PR actually break things. it 
- uses the old `apt:key` syntax
- ensure different packages as it require afterwards
- uses the old facts syntax
- hardcodes the repo key id
- cannot be disabled (something like `$sury = false`)

EpiqSty:
Hi,

First of all thanks for great work ;)  We used that module for a while, but now we have tried to install PHP 7.1 on Debian 9 , and faced with those errors:
```
Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install php7.1-cli' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package php7.1-cli
E: Couldn't find any package by glob 'php7.1-cli'
E: Couldn't find any package by regex 'php7.1-cli'
Error: /Stage[main]/Php::Packages/Package[php7.1-cli]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install php7.1-cli' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package php7.1-cli
E: Couldn't find any package by glob 'php7.1-cli'
E: Couldn't find any package by regex 'php7.1-cli'
Error: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install php7.1-common' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package php7.1-common
E: Couldn't find any package by glob 'php7.1-common'
E: Couldn't find any package by regex 'php7.1-common'
Error: /Stage[main]/Php::Packages/Package[php7.1-common]/ensure: change from 'purged' to 'present' failed: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold install php7.1-common' returned 100: Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package php7.1-common
E: Couldn't find any package by glob 'php7.1-common'
E: Couldn't find any package by regex 'php7.1-common'
```
Does it mean, what such combination (PHP7.1 & Debian 9) still not supported by module, or we have to add some custom configuration?


For now we just defined desired version this way:
```
php::globals::php_version: '7.1'
```

is it enough or not?

Issue #267, opened 2016-11-05
repo::debian uses deprecated options on puppetlabs-apt


kwisatz:
The following notifications are being issues when using this module on a Debian host:

```
Nov 05 19:59:30 puppet puppet-master[21829]: (Scope(Apt::Key[php::repo::debian])) $key is deprecated and will be removed in the next major release. Please use $id instead.
Nov 05 19:59:30 puppet puppet-master[21829]: (Scope(Apt::Key[php::repo::debian])) $key_source is deprecated and will be removed in the next major release. Please use $source instead.
Nov 05 19:59:30 puppet puppet-master[21829]: (Scope(Apt::Source[source_php_jessie])) $include_src is deprecated and will be removed in the next major release, please use $include => { 'src' => false } instead
```

PR #296, opened 2017-01-19
update deprecated apt::key parameters

cbaoxa:


cbaoxa:
fixes #267

hunner:
Looks like the metadata already has apt > 2.1.0 so this is 👍

PR #297, opened 2017-01-19
allow pipe param for pecl extensions

cbaoxa:
cf https://github.com/voxpupuli/puppet-php/blob/master/lib/puppet/provider/package/pecl.rb#L83

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -52,6 +52,10 @@
 #   String parameter, whether to specify ALL sapi or a specific sapi.
 #   Defaults to ALL.
 #
+# [*pipe*]
+#   String parameter to input answers during extension setup. Supported
+#   *provider*: pecl.
+#
 define php::extension (
   $ensure            = 'installed',
   $provider          = undef,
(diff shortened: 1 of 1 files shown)


manifests/extension.pp
```diff
@@ -66,6 +70,7 @@
   $settings          = {},
   $settings_prefix   = false,
   $sapi              = 'ALL',
+  $pipe              = undef,
```
alexjfisher: Does `$pipe` then not also have to be passed as an argument to `ensure_packages`?

cbaoxa: Thanks! I ran the wrong test on my side and missed this. PR updated

manifests/extension.pp
```diff
@@ -52,6 +52,10 @@
 #   String parameter, whether to specify ALL sapi or a specific sapi.
 #   Defaults to ALL.
 #
+# [*pipe*]
+#   String parameter that may be used by the *provider* to input answers during
```
alexjfisher: Maybe mention it's for the 'pecl' provider only here?

cbaoxa: Hoping this will be updated correctly in the future, sure :)

Issue #303, opened 2017-02-11
composer's self-update fails, missing environment variables


igalic:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 4.9.1
- Ruby: 2.1.9
- Distribution: Ubuntu trusty
- Module version: 03e1187f6585ef9c09724f1461c60f51790ff236

## How to reproduce (e.g Puppet code you use)

```puppet
include php
```

## What are you seeing

```
Notice: Compiled catalog for tapp03.esat in environment production in 7.56 seconds
Notice: /Stage[main]/Php::Composer::Auto_update/Exec[update composer]/returns: 
Notice: /Stage[main]/Php::Composer::Auto_update/Exec[update composer]/returns:                                                                                             
Notice: /Stage[main]/Php::Composer::Auto_update/Exec[update composer]/returns:   [RuntimeException]                                                                        
Notice: /Stage[main]/Php::Composer::Auto_update/Exec[update composer]/returns:   The HOME or COMPOSER_HOME environment variable must be set for composer to run correctly  
Notice: /Stage[main]/Php::Composer::Auto_update/Exec[update composer]/returns:                                                                                             
Error: /usr/local/bin/composer --no-interaction --quiet self-update returned 1 instead of one of [0]
Error: /Stage[main]/Php::Composer::Auto_update/Exec[update composer]/returns: change from notrun to 0 failed: /usr/local/bin/composer --no-interaction --quiet self-update returned 1 instead of one of [0]
```
## What behaviour did you expect instead

I would've expected this to _just work™_

## Any additional information you'd like to impart

I've speculated that rather than using `composer self-update`, we should instead use extend archive: https://github.com/voxpupuli/puppet-archive/issues/244 since redownloading seems to be a common usecase

igalic:
if my understanding is correct, we should add a default `{ HOME => '/root' }` to the `$env` here: https://github.com/voxpupuli/puppet-php/blob/master/manifests/composer/auto_update.pp#L40-L44

PR #305, opened 2017-02-11
composer auto_update: exec's environment must be array

igalic:
we were passing a Hash, a Hash missing HOME, at that (#303)
This fixes both issues.


bastelfreak:
ruby19 issues will be handled in https://github.com/voxpupuli/puppet-php/pull/304

PR #308, opened 2017-02-13
Remove support for PHP 5.5 in Ubuntu

thaiphv:
PHP 5.5 on Ubuntu is no longer available for download from the "ondrej/php" PPA. I think we should fail the execution when people want to install it.

### DIFF
--- a/manifests/repo/ubuntu.pp
+++ b/manifests/repo/ubuntu.pp
@@ -16,11 +16,14 @@ class php::repo::ubuntu (
     $version_real = $version
   }
 
+  if ($version_real == '5.5') {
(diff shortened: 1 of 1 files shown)


igalic:
@thaiphv thank you very much for this pr
you may want to consider adding the commit email to your github profile (for internet points)
or else, amending this commit's author

thaiphv:
@igalic thanks for the suggestions. I've just pushed my email change.

sacres:
Apologies for not saying so when merging/reviewing: Thanks @thaiphv  !

thaiphv:
@sacres no worries. Thanks for accepting it.

PR #310, opened 2017-02-17
no further config required for PEAR extensions

mcrauwel:
the current behaviour will generate the following output on every php-cli action (including every cronjob)

```
Warning: Could not match PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php/20131226/Image_Canvas.so' - /usr/lib/php/20131226/Image_Canvas.so: cannot open shared object file: No such file or directory in Unknown on line 0
Warning: Could not match PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php/20131226/Image_Color.so' - /usr/lib/php/20131226/Image_Color.so: cannot open shared object file: No such file or directory in Unknown on line 0
Warning: Could not match PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php/20131226/Image_Graph.so' - /usr/lib/php/20131226/Image_Graph.so: cannot open shared object file: No such file or directory in Unknown on line 0
```

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -142,74 +142,76 @@ define php::extension (
     $package_depends = undef
   }
 
-  if $zend == true {
-    $extension_key = 'zend_extension'
-    if $php_api_version != undef {
-      $module_path = "/usr/lib/php5/${php_api_version}/"
+  if $provider != 'pear' { # PEAR packages don't require antu further configuration, they just need to "be there".
+    if $zend == true {
+      $extension_key = 'zend_extension'
+      if $php_api_version != undef {
+        $module_path = "/usr/lib/php5/${php_api_version}/"
+      }
+      else {
+        $module_path = undef
+      }
     }
     else {
+      $extension_key = 'extension'
       $module_path = undef
     }
-  }
-  else {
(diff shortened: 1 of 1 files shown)


manifests/extension.pp
```diff
@@ -142,74 +142,76 @@
...
 
-  if $zend == true {
-    $extension_key = 'zend_extension'
-    if $php_api_version != undef {
-      $module_path = "/usr/lib/php5/${php_api_version}/"
+  if $provider != 'pear' { # PEAR packages don't require antu further configuration, they just need to "be there".
+    if $zend == true {
+      $extension_key = 'zend_extension'
+      if $php_api_version != undef {
+        $module_path = "/usr/lib/php5/${php_api_version}/"
```
igalic: what about versions of PHP that are *not* 5?

mcrauwel: basicly, it wouldn't have worked before... I guess...
I'll look into this, but maybe we should do it in a seperate pull-req since it exceeds the scope of this one?

igalic:
kaythnkx

PR #313, opened 2017-02-23
Simplify php::extension

igalic:
This defined type tried to do too much at once, so split the two main
concerns: installation & onfiguration.

Readability is greatly enhanced by reducing the nesting level of `if`s.

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -18,10 +18,6 @@
 #   The source to install the extension from. Possible values
 #   depend on the *provider* used
 #
-# [*pecl_source*]
-#   The pecl source channel to install pecl package from
-#   Superseded by *source*
-#
 # [*so_name*]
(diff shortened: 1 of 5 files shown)


igalic:
something in our spec tests for RHEL is broken, and i don't know what

manifests/extension.pp
```diff
@@ -57,161 +53,46 @@
...
-  } else {
-    $package_depends = undef
+  php::extension::install { $title:
+    ensure            => $ensure,
+    provider          => $provider,
+    source            => $source,
+    responsefile      => $responsefile,
+    header_packages   => $header_packages,
+    compiler_packages => $compiler_packages,
   }
```
yastupin: failing tests are due to package_prefix not being passed to extension::install

tested locally, rake spec is fine on centos6/7 and rhel6/7

igalic: oops, thanks for catching this

PR #322, opened 2017-03-06
fixes for mysql-extension on php7 and fixes for repos on debian

mcrauwel:

### DIFF
--- a/manifests/extension/config.pp
+++ b/manifests/extension/config.pp
@@ -76,10 +76,17 @@ define php::extension::config (
     String => ensure_prefix($settings, "${settings_prefix}."),
   }
 
-  $final_settings = deep_merge(
-    {"${extension_key}" => "${module_path}${so_name}.so"},
-    $full_settings
-  )
+  case $so_name {
+    'mysql': {
+      $final_settings = $full_settings
+    }
+    default: {
+      $final_settings = deep_merge(
+        {"${extension_key}" => "${module_path}${so_name}.so"},
+        $full_settings
+      )
+    }
+  }
 
   $config_root_ini = pick_default($::php::config_root_ini, $::php::params::config_root_ini)
   ::php::config { $title:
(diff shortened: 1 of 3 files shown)


bastelfreak:
looks fine to me. Can somebody else review it as well? I'm not very familiar with the internals of the php module.

manifests/extension/config.pp
```diff
@@ -76,10 +76,17 @@
     String => ensure_prefix($settings, "${settings_prefix}."),
   }
 
-  $final_settings = deep_merge(
-    {"${extension_key}" => "${module_path}${so_name}.so"},
-    $full_settings
-  )
+  case $so_name {
```
ffrank: I'm getting a little dizzy trying to decipher this whole logic, including the `$full_settings` bit above.

Could you add some comments to make this more clear?

ffrank:
Apart from on inline comment I made, this looks good to me as well (same caveat though).

I'm willing to merge this once I have some feedback regarding my request for comments (pun intended I guess?)

manifests/extension/config.pp
mcrauwel: hi @ffrank,

I've had to revisit the code myself to figure out what it was doing...

This define gets a `$settings`-Hash as parameter. If you have it configured to prefix the settings then it will do so resulting in the `$full_settings`-Hash. 

The previous behaviour would always a setting: 
```
extension = /path/to/so_name.so
```

This doesn't work anymore with the MySQL-module in 7.x since the `mysql` extension was removed in favour of `mysqli`, `mysqlnd` or `pdo_mysql`. Which is the second block of code to enable those modules...

If needed I will add more comment tonight in the code.

ffrank: @mcrauwel sorry for the late reply.
Yes, please add some code comments.

bastelfreak:
@mcrauwel can you please rebase?

c33s:
@bastelfreak 
as i need this code too, i can make a new PR based on the master and add merge the changes. also i would do suggest making seperate PR out of this one. one for the mysql fix which is the most important and a 2nd one for the better repo handling.

bastelfreak:
@c33s sounds good!

bastelfreak:
Thanks for opening this issue. It hasn't seen activity in a while, so I'm going to close the issue. This doesn't mean that we don't think this issue is valid, we're just trying to assess what issues are still relevant. If you still need this(even if you're not willing to work on it yourself) please feel free to reopen it. Again thanks for opening the issue, users who report bugs and ask for features are some of the most important players in the open source community. Feel free to hop by #voxpupuli on irc.freenode.net if you'd like to chat with us about any of this.

PR #325, opened 2017-03-10
Skip management of MySQL PHP module .ini file if using Ubuntu and repo

tdm4:
The Ondřej Surý PHP repo automatically configures the module .ini files. There is no need for php::extension::config to do anything. PR #322  only applies this to the mysql module, but, the PHP .deb packages already configure their modules in post-install.

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '0.20.0'
+modulesync_config_version: '0.20.1'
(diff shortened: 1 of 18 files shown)


tdm4:
Not sure about these travis-ci errors:

should contain Package[php5-json] (FAILED - 1)
Isn't that the wrongly named package if you're using the Ondřej Surý Ubuntu repo?

PR #329, opened 2017-03-16
Allow installing PHP 7.1 on Debian

amette:
Hi,

this simple patch made installation of PHP 7.1 work for me on Debian 8 Jessie with manually configured  https://deb.sury.org/ repositories.

cheers
amette


manifests/globals.pp
```diff
@@ -83,7 +83,7 @@
             $fpm_service_name     = "php${globals_php_version}-fpm"
             $ext_tool_enable      = "/usr/sbin/phpenmod -v ${globals_php_version}"
             $ext_tool_query       = "/usr/sbin/phpquery -v ${globals_php_version}"
-            $package_prefix       = 'php7.0-'
+            $package_prefix       = "php${globals_php_version}-"
```
yastupin: This will break for php 5, where $default_php_version is "5.x" and no just "5" :/

yastupin:
Oh right, sorry I read the patch wrong. Thank you!

PR #331, opened 2017-03-17
require pear in pecl file

jarro2783:
This fixes the `pecl` provider autoloading the `pear` provider. On Puppet 4.9.4, I get the following error:
```
     Puppet::Error:
       Could not autoload puppet/type/package: Could not autoload puppet/provider/package/pecl: cannot load such file -- puppet/provider/pear
```

It doesn't seem to happen on Puppet 3. This is fixed by requiring `pear` explicitly in the `pecl` provider code.

### DIFF
--- a/lib/puppet/provider/package/pecl.rb
+++ b/lib/puppet/provider/package/pecl.rb
@@ -1,5 +1,5 @@
 require 'puppet/provider/package'
-require 'puppet/provider/package/pear'
+require_relative './pear'
 
 Puppet::Type.type(:package).provide :pecl, parent: :pear do
   desc 'Package management via `pecl`.'
--- a/lib/puppet/provider/package/pecl.rb
+++ b/lib/puppet/provider/package/pecl.rb
@@ -1,4 +1,5 @@
 require 'puppet/provider/package'
+require 'puppet/provider/package/pear'
 
 Puppet::Type.type(:package).provide :pecl, parent: :pear do
   desc 'Package management via `pecl`.'


jarro2783:
It looks like my original fix might not actually work, even though it solved the problem on my dev host, it then couldn't find `puppet/provider/package/pear` on our CI host. I'm now doing `require_relative './pear'`.

yastupin:
Could not reproduce with puppet 4.10.0 (agent and master) and the following manifest in /etc/puppetlabs/code/environments/production/manifests/php.pp:

```
include apt

class { 'php': }

php::extension { 'event':
  provider => 'pecl',
}
```

The extension is installed correctly:
```
root@a75003b4bbfb:/# pecl list
Installed packages, channel pecl.php.net:
=========================================
Package Version State
event   2.3.0   stable
```

Looks like an autoload issue on your side. Are your modules in the standard path, under production (or current) environment?

jarro2783:
At the moment it seems like it only causes problems when I use `rspec-puppet`. The other strange thing is that it only fails for the first test. Subsequent tests seem to have loaded the file correctly.

I don't know if it's actually related to the error that it prints before that. Here is some context:
```
Failure/Error: Puppet::Type.type(:package).stubs(:defaultprovider).returns(Puppet::Type.type(:package).provider(:apt))

     Puppet::Error:
       Could not autoload puppet/type/package: Could not autoload puppet/provider/package/pecl: Could not find parent provider pear of pecl
```

ffrank:
What's the exact reproduction? Do you just invoke `rake spec` on current master?

jarro2783:
It does it for tests under `spec/hosts`, and I just run `rake spec`.

Just to be clear, this is running in our codebase, not this module. I wonder if it's actually a problem with `rspec-puppet`.

ffrank:
Can you please clarify? Are you saying that this module trips up rspec-puppet running the custom tests in your own codebase? Is this module pulled as a dependency there?

jarro2783:
Yes, the problem happens running our own tests where this module is a dependency, and only for [hosts](http://rspec-puppet.com/documentation/hosts/) tests. So I'm actually wondering if this is a problem with `rspec-puppet` and not this module.

ffrank:
If you can create a clone of your environment and strip it down to the bare essentials that will still demonstrate this issue, that would be very helpful to have.

bastelfreak:
@jarro2783 can you please rebase against our latest master? We added acceptance tests and should now be able to reproduce this.

yakatz:
Not clear to me if the tests ran. I suggest also rebasing and squashing the commits (which should retrigger the testing anyway).

yakatz:
This PR has had no activity in a long time. Feel free to rebase/update/resubmit if this is still relevant.

PR #333, opened 2017-03-24
Add optional param to prefix the extension ini filename

spacepants:
Hey there,

So when loading extensions, sometimes it's desirable to specify the extension load order — for example loading opcache before other modules. This adds support for prefixing the extension ini filename to accomplish this.

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -21,6 +21,9 @@
 # [*so_name*]
 #   The DSO name of the package (e.g. opcache for zendopcache)
 #
+# [*ini_prefix*]
+#   An optional filename prefix for the settings file of the extension
+#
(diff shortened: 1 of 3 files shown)


jk2l:
how will this fix the duplicate issue? let say if the default installed 10-module.ini and you want to raise the load order to 20-module.ini, the issue same with other similar pull request it will not delete the 10-module.ini caused duplicate ini file.

poikilotherm:
@jk2l, you are right. At least, you don't get duplicated files with default names like "10-module.ini" and "module.ini". Instead, the default file will get edited and options are added.

Maybe @spacepants could extend his code to use a `tidy` resource to cleanup any unmanaged files for the extensions now under puppet control?

PR #337, opened 2017-04-10
Feature/pm ondemand

derkgort:
Hi,

This adds the option to use the ondemand pm.

### DIFF
--- a/manifests/fpm/pool.pp
+++ b/manifests/fpm/pool.pp
@@ -42,6 +42,8 @@
 #
 # [*pm_max_requests*]
 #
+# [*pm_process_idle_timeout*]
+#
 # [*pm_status_path*]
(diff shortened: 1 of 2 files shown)


derkgort:
These fails are not in my added code...

yastupin:
Looks like you're adding the `pm_process_idle_timeout` option here.
Is this the expected commit?

derkgort:
Yes.
quoting the php-fpm documentation:

> pm.process_idle_timeout mixed
The number of seconds after which an idle process will be killed. Used only when pm is set to ondemand. Available units: s(econds)(default), m(inutes), h(ours), or d(ays). Default value: 10s.

yastupin:
Ok thanks for the explanation!

PR #347, opened 2017-06-02
Restrict Apt test fixture version to module dependency

dbeckham:
This change fixes currently failing tests by restricting the puppetlabs-apt module version in `.fixtures.yml` to the latest version actually supported and defined in `metadata.json`.  

Long term, all modules included in the fixtures file should have versions defined so that the current `apt::source` parameter problem doesn't happen in the future.

Fwiw, this isn't meant to be a fix for the changes to apt::source like PR #342 is meant to be, but to fix a testing bug.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,7 +1,9 @@
 fixtures:
   repositories:
     stdlib: "git://github.com/puppetlabs/puppetlabs-stdlib.git"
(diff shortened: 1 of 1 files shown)


yastupin:
FYI tests were fixed in #349

bastelfreak:
Hi @dbeckham, thanks for the PR. I think this is obsolete now.

PR #350, opened 2017-06-12
replace validate_* with puppet4 datatypes

bastelfreak:


bastelfreak:
this should be rebased after https://github.com/voxpupuli/puppet-php/pull/349 got merged

bastelfreak:
I adjusted the types, there should be no backwards-incompatible changes.

manifests/composer.pp
```diff
@@ -24,24 +24,19 @@
...
-  $auto_update  = true,
-  $max_age      = $::php::params::composer_max_age,
-  $root_group   = $::php::params::root_group,
+  String $source             = $::php::params::composer_source,
+  Stdlib::Absolutepath $path = $::php::params::composer_path,
+  $proxy_type                = undef,
+  $proxy_server              = undef,
+  Boolean $auto_update       = true,
+  Integer $max_age           = $::php::params::composer_max_age,
+  $root_group                = $::php::params::root_group,
```
rnelson0: `Variant[String, Integer]`?

manifests/config/setting.pp
```diff
@@ -21,15 +21,13 @@
 define php::config::setting(
   $key,
   $value,
-  $file,
+  String $file,
```
rnelson0: This should probably be a `Stdlib::AbsolutePath` since it is passed to a `path` parameter.

manifests/embedded.pp
```diff
@@ -15,20 +15,17 @@
 #   Specify which version of the package to install
 #
 class php::embedded(
-  $ensure   = $::php::ensure,
-  $package  =
+  $ensure                       = $::php::ensure,
+  $package                      =
```
rnelson0: Shouldn't these be strings as in `php::dev`?

manifests/composer.pp
bastelfreak: yep

manifests/embedded.pp
bastelfreak: yep

manifests/config/setting.pp
bastelfreak: yep

manifests/composer.pp
```diff
@@ -24,24 +24,19 @@
...
-  $auto_update  = true,
-  $max_age      = $::php::params::composer_max_age,
-  $root_group   = $::php::params::root_group,
+  String $source             = $::php::params::composer_source,
+  Stdlib::Absolutepath $path = $::php::params::composer_path,
+  $proxy_type                = undef,
+  $proxy_server              = undef,
+  Boolean $auto_update       = true,
+  Integer $max_age           = $::php::params::composer_max_age,
+  Variant[Integer, String] $root_group                = $::php::params::root_group,
```
rnelson0: Alignment is off here (I don't care but I think our standard does).

manifests/globals.pp
```diff
@@ -12,20 +12,10 @@
 #   Path to pid file for fpm
 
 class php::globals (
-  $php_version  = undef,
-  $config_root  = undef,
-  $fpm_pid_file = undef,
+  Optional[Pattern[/^[57].[0-9]/]] $php_version  = undef,
+  Optional[Stdlib::Absolutepath] $config_root  = undef,
+  Optional[Stdlib::Absolutepath] $fpm_pid_file = undef,
```
rnelson0: Alignment is off.

manifests/packages.pp
```diff
@@ -13,22 +13,18 @@
 #   package prefix `$php::package_prefix`
 #
 class php::packages (
-  $ensure          = $::php::ensure,
-  $manage_repos    = $::php::manage_repos,
-  $names_to_prefix = prefix(
+  String $ensure         = $::php::ensure,
+  Boolean $manage_repos  = $::php::manage_repos,
+  Array $names_to_prefix = prefix(
     $::php::params::common_package_suffixes, $::php::package_prefix # lint:ignore:parameter_documentation
```
rnelson0: We should be consistent about how we merge two values for a parameter name. Compare to `manifests/embedded.pp` L18 or so.

manifests/composer.pp
bastelfreak: fixed

manifests/globals.pp
bastelfreak: fixed

dbeckham:
The changes for `pool.pp` (abee5c17fd95bf1990fe95904ff9dd8dad7d19fd) are missing all of the Puppet 4 Data Type definitions except for one.

bastelfreak:
@dbeckham this is actually intentional. The goal for me is to get rid of all the legacy validate_* calls, because they are spamming the puppetserver logs and they get removed in the future. My goal (at the moment) is not to provide perfect types for all parameter, I don't have enough for it at the moment. However, we're happily accepting patches :)

PR #351, opened 2017-06-12
modulesync 0.21.3

bastelfreak:


bastelfreak:
this should be rebased after https://github.com/voxpupuli/puppet-php/pull/349 got merged

PR #357, opened 2017-06-29
replace legacy facts with new $facts hash

bastelfreak:

### DIFF
--- a/manifests/fpm.pp
+++ b/manifests/fpm.pp
@@ -68,7 +68,7 @@ class php::fpm (
 
   # On FreeBSD fpm is not a separate package, but included in the 'php' package.
   # Implies that the option SET+=FPM was set when building the port.
-  $real_package = $::osfamily ? {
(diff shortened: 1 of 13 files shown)


manifests/fpm.pp
```diff
@@ -98,7 +98,7 @@
 
   # Create an override to use a reload signal as trusty and utopic's
   # upstart version supports this
-  if $::operatingsystem == 'Ubuntu' and ($::operatingsystemmajrelease == '14.04' or $::operatingsystemmajrelease == '14.10') {
+  if $facts['os']['name'] == 'Ubuntu' and ($facts['os']['release']['major'] == '14') {
```
alexjfisher: Turns out 'major' is actually '14.04' not just '14'. https://github.com/camptocamp/facterdb/blob/54422fbe9e32cfacddae5403237ee2758a7f26af/facts/3.7/ubuntu-14.04-x86_64.facts#L339

PR #361, opened 2017-07-04
fix default value of php::fpm::pool::access_log_format

lesinigo:
The default value for `php::fpm::pool::access_log_format` gets its escapes un-escaped by Puppet string management, resulting in a log line that looks like this:
```
127.0.0.1 -  04/Jul/2017:11:33:53 +0200GET /test.php404
```

this PR fixes the default value and gives you correctly formatted log lines like:
```
127.0.0.1 -  04/Jul/2017:12:36:24 +0200 "GET /test.php" 404
```


bastelfreak:
oh annoying bug, thanks for the fix @lesinigo!

Issue #372, opened 2017-08-09
Changes in #357 break Ubuntu version dependent resources


dbeckham:
Changes in #357 break Ubuntu version dependent resources when using Facter 3.x.

The changes integrate the new facts hash from Puppet 4, but it appears that the version strings were also changed to `12`, `14`, etc., which is broken for current releases of Puppet 4 and Facter 3.x.  Facter reports these facts as `12.04`, `14.04`, or `14.10`, etc.  The old Facter 2.x was reporting `12` and `14`, but it appears that it can be very buggy from release to release.

This behavior can be verified by installing Puppet 4 from the [official Puppetlabs Ubuntu repository](https://docs.puppet.com/puppet/3.8/puppet_repositories.html#apt-based-systems).

Either way, using  `$foo == '14.04'`, or `$foo == '14'` are both short sighted and will fail for anything but those exact versions. What should be done, and in general a Puppet best practice, is to use the built-in [versioncmp](https://docs.puppet.com/puppet/latest/function.html#versioncmp) function to do a version comparison to ensure that your resources are applied to any applicable versions.

Examples:

```puppet
# Ubuntu 14 only
if (versioncmp($facts['os']['release']['full'], '14') >= 0 and 
    versioncmp($facts['os']['release']['full'], '16') < 0) {
  # do something version specific
}

# Ubuntu 16 and later releases:
if versioncmp($facts['os']['release']['full'], '16') >= 0 {
  # do something version specific
}


```

dbeckham:
Well, I looked over #357 and saw a couple of `==` changes for the Ubuntu version and made an assumption about all of the changes. My apologies. This doesn't affect ALL ubuntu versions, just 12 and 14. I'm going to submit a PR with a fix in it for the affected manifests.

PR #375, opened 2017-08-10
Fix OS fact comparison for Ubuntu 12 and 14

dbeckham:
This change uses `versioncmp` instead of an equality operator when
comparing Ubuntu OS facts for more compatibility between various facter
versions, and Ubuntu system versions.

Also, `$facts['os']['release']['major']` was changed to
`$facts['os']['release']['full']` to be more consistent with other
version comparisons in this module.

Fixes the issue reported in #372.


alexjfisher:
@dbeckham Thanks!

PR #381, opened 2017-09-15
modulesync 1.1.0

bastelfreak:

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -1,1 +1,1 @@
-modulesync_config_version: '0.21.3'
+modulesync_config_version: '1.1.0'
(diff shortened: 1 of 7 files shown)


bastelfreak:
any rspec experts around? I'm not able to reproduce this locally with puppet5.2 and ruby2.4

alexjfisher:
Failure isn't puppet 5 specific.  Also fails on puppet 4 with ruby 2.2.

wyardley:
@bastelfreak @alexjfisher fixed in #386 - not at all an expert, but this was similar to my fix for nodejs module.

PR #385, opened 2017-09-26
Fix syntax issues with data types

craigwatson:
```
Error: Syntax error at ']' at /etc/puppetlabs/code/modules/php/types/sapi.pp:6:1
Error: Syntax error at ']' at /etc/puppetlabs/code/modules/php/types/provider.pp:28:1
```


alexjfisher:
@craigwatson Thanks!

PR #408, opened 2018-01-27
fix composer auto_update idempotency in case no update is available

### LINKED ISSUE #402: Auto_update not idempotent

- Puppet: 4.10.x
- Ruby: 2.1.9 (via `facter rubyversion`)
- Distribution: centos7x
- Module version: 5.1.0 (https://forge.puppet.com/puppet/php)

```puppet
    class { '::php':
        settings   => {
            'PHP/max_execution_time'  => $max_execution,
            'PHP/max_input_time'      => $max_input,
            'PHP/memory_limit'        => $memory_limit,
            'PHP/post_max_size'       => $post_max,
            'PHP/upload_max_filesize' => $upload_max,
            'Date/date.timezone'      => $timezone,
        },
        manage_repos => true,
        proxy_type   => 'https',
        proxy_server => $my_proxy,
        composer     => true,
        extensions   => {
            pdo       => { },
            pdo_mysql => { },
        }
    }
```

## What are you seeing

The module is not idempotent, and I keep seeing:

```bash
Notice: /Stage[main]/Php::Composer::Auto_update/Exec[update composer]/returns: executed successfully
```

I expected the module to be idempotent, with respect to composer.

## Any additional information you'd like to impart

If possible, I suggest making the following logic from [`composer.pp`](https://github.com/voxpupuli/puppet-php/blob/f44104319cfd089599850bb9eb2fa9f2afd784ed/manifests/composer.pp#L52-L59) idempotent, by maybe creating a fact, to check if there is an update available, then run `auto_update` (otherwise, don't run):

```puppet
  if $auto_update {
    class { '::php::composer::auto_update':
      max_age      => $max_age,
      source       => $source,
      path         => $path,
      proxy_type   => $proxy_type,
      proxy_server => $proxy_server,
    }
}
}
```

If my suggestion seems reasonable, I could potentially create the puppet fact, and integrate it, to the above `class`?  For the time being, I'm going to try to set `$auto_update` to `false`.

joekohlsdorf:
This fixes #402.

In case ```max_age``` has passed but no update was available we attempted to update on every run.

### DIFF
--- a/manifests/composer/auto_update.pp
+++ b/manifests/composer/auto_update.pp
@@ -44,6 +44,7 @@ class php::composer::auto_update (
   }
 
   exec { 'update composer':
+    # touch binary when an update is attempted to update its mtime for idempotency when no update is available
     command     => "${path} --no-interaction --quiet self-update; touch ${path}",
     environment => $env,
     onlyif      => "test `find '${path}' -mtime +${max_age}`",
--- a/manifests/composer/auto_update.pp
+++ b/manifests/composer/auto_update.pp
@@ -44,7 +44,7 @@ class php::composer::auto_update (
   }
 
   exec { 'update composer':
-    command     => "${path} --no-interaction --quiet self-update",
+    command     => "${path} --no-interaction --quiet self-update; touch ${path}",
     environment => $env,
     onlyif      => "test `find '${path}' -mtime +${max_age}`",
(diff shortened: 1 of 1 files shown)


bastelfreak:
Thanks for the PR @joekohlsdorf !

PR #409, opened 2018-01-28
Harden default php settings

juju4:
Following https://github.com/puppetlabs/puppetlabs-apache/pull/1749#issuecomment-358125758, I adapted to puppet-php.
For now it is just part of settings, not sure if it should be differentiated.
Risk: users silently overwriting it vs potential post-adaption needed

Potentially more settings to add.


c33s:
this pr is a bc break, so if this module should follow semver, it would require a major version number increase.

of course i am for good defaults, but it is more than annoying if a minor or patch update break things.

for example `expose_php` is quite harmless but settings like `allow_url_open` changed in a minor or patch version are fatal.

also the settings which are set to a hardened default should be discussed

c33s:
my suggestion would be to add a boolean flag named `harden` and only if this is set to `true` the harden settings will be applied (and can still by overriden by providing settings in the `settings` hash)

manifests/init.pp
```diff
@@ -136,7 +136,20 @@
   $proxy_type                                     = undef,
   $proxy_server                                   = undef,
   Hash $extensions                                = {},
-  Hash $settings                                  = {},
+  Hash $settings                                  = {
```
bastelfreak: I think we need at least one acceptance test to verify that those settings actually work.

bastelfreak:
I agree with @c33s here. I'm fine with breaking changes as long as we make it in a major release (which is also fine). We always try to ship working default settings that are as close as the upstream settings. Hardened settings are good, but they are often opinionated.  I'm happy to merge this but it needs to be optional and we need tests. We can't ship settings when we don't know if they work.

juju4:
I'm fine with current remarks. can have it in tree with harden false and at next major release, set to true.

few questions
- if I differentiate those settings (eg harden_enable, harden_settings), is there a puppet function to merge settings and harden_settings with priority to the former?
- as for test, adding a harden_spec.rb in spec/classes seems the way to go. correct?

bastelfreak:
@juju4 unit tests go into spec/classes/, acceptance tests into spec/acceptance/. There are multiple solutions for merging hashes, for example https://forge.puppet.com/puppetlabs/stdlib#deep_merge

juju4:
Sorry, it's on my todo list but had no time for it for now. maybe in a few weeks.

juju4:
On default values, there are the ones I used and would recommend.
There are obviously very context dependent but IMO, I prefer a secure default state with known documentation and easy to adapt that the often unsecure unknown default.

Some are from audit tools or other.
Feel free to adapt

manifests/init.pp
```diff
@@ -147,7 +147,20 @@
...
+    # php secure settings by default
+    'Date/date.timezone'      => 'UTC',
+    'PHP/expose_php'          => 'Off',
+    'PHP/allow_url_fopen'     => 'Off',
+    'PHP/disable_functions'   => 'pcntl_alarm,pcntl_fork,pcntl_waitpid,pcntl_wait,pcntl_wifexited,pcntl_wifstopped,pcntl_wifsignaled,pcntl_wexitstatus,pcntl_wtermsig,pcntl_wstopsig,pcntl_signal,pcntl_signal_dispatch,pcntl_get_last_error,pcntl_strerror,pcntl_sigprocmask,pcntl_sigwaitinfo,pcntl_sigtimedwait,pcntl_exec,pcntl_getpriority,pcntl_setpriority,chown,diskfreespace,disk_free_space,disk_total_space,dl,exec,escapeshellarg,escapeshellcmd,fileinode,highlight_file,max_execution_time,passthru,pclose,phpinfo,popen,proc_close,proc_open,proc_get_status,proc_nice,proc_open,proc_terminate,set_time_limit,shell_exec,show_source,system,serialize,unserialize,__construct, __destruct, __call,__wakeup',
```
igalic: aaaand, before 7.0.0, you may as well add [`preg_replace`](https://www.php.net/manual/en/reference.pcre.pattern.modifiers.php#reference.pcre.pattern.modifiers.eval)

yakatz:
This PR has not had any reply to the reviews and has been open for a long time with no other activity. There is also no consensus on making such breaking changes. Feel free to revise/update/resubmit, but closing for now.

PR #410, opened 2018-01-31
Deprecate hiera_hash functions

minorOffense:
Having some merge issues with hiera_hash in later versions of puppet. The deep merge isn't working as expected. However using lookup instead fixes the issue. And since hiera_hash will be deprecated anyways I figure start converting some of those calls.


bastelfreak:
Thanks for this @minorOffense !

PR #412, opened 2018-02-03
add ubuntu 16.04 support

bastelfreak:
This was mostly working, it just wasn't tested. I added the tests and fixed some edge cases.

### DIFF
--- a/manifests/dev.pp
+++ b/manifests/dev.pp
@@ -9,8 +9,9 @@
 #   The package name for the PHP development files
 #
 class php::dev(
-  String $ensure  = $::php::ensure,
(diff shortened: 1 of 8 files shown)


manifests/dev.pp
```diff
@@ -23,21 +24,31 @@
...
     ensure_packages(["${php::package_prefix}xml"], {
       ensure  => present,
-      require => Class['::apt::update'],
+      require => $require,
     })
 
     package { $real_package:
       ensure  => $ensure,
-      require => Class['::php::packages'],
+      require => $require,
```
alexjfisher: Are you sure this is correct?

PR #420, opened 2018-02-23
Fix wrongly named parameters for apt::source/key

bitcrush:
This PR fixes a naming issue in php::repo::debian. The parameters given to the apt::key and apt::source defined types are not valid for the puppetlabs/apt module >= 4.4.0, which this module depends on.

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -20,6 +20,9 @@
 # [*dotdeb*]
 #   Enable special dotdeb handling
 #
+# [*sury*]
+#   Enable special sury handling
+#
 class php::repo::debian(
   $location     = 'http://packages.dotdeb.org',
   $release      = 'wheezy-php56',
@@ -30,6 +33,7 @@ class php::repo::debian(
     'source' => 'http://www.dotdeb.org/dotdeb.gpg',
   },
   $dotdeb       = true,
+  $sury         = true,
 ) {
 
   if $caller_module_name != $module_name {
@@ -70,7 +74,7 @@ class php::repo::debian(
     }
   }
 
-  if ($php::globals::php_version == '7.1') {
+  if ($sury and $php::globals::php_version == '7.1') {
     # Required packages for PHP 7.1 repository
     ensure_packages(['lsb-release', 'ca-certificates'], {'ensure' => 'present'})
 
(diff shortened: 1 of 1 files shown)


c33s:
can you please also put the whole  php7.1 part in an if and add a flag to it?

```
class php::repo::debian(
  $location     = 'http://packages.dotdeb.org',
  $release      = 'wheezy-php56',
  $repos        = 'all',
  $include_src  = false,
  $key          = {
    'id'     => '6572BBEF1B5FF28B28B706837E3F070089DF5277',
    'source' => 'http://www.dotdeb.org/dotdeb.gpg',
  },
  $dotdeb       = true,
  $sury         = true,
) {
```

and

```

if ($sury and $php::globals::php_version == '7.1') {
...
```
as i wrote https://github.com/voxpupuli/puppet-php/pull/293#issuecomment-366333850 if you use the manage repo and provide the sury repo by hiera the new php7.1 section in this module part break it.

bitcrush:
I implemented the requested flag. However the class lags functionality/configurabiltity. [This](https://gist.github.com/cea5b356c9e0dff3e48acaad3bc23db7) is a proposal that should solve the issues mentioned in [#293](https://github.com/voxpupuli/puppet-php/pull/293#issuecomment-366333850). In addition it adds support for PHP 7.2 on Jessie and Stretch. Please have a look and let me know if I should create another PR for it.

c33s:
if i would have merge permissions i would merge/approve your PR but we have to wait for someone with merge permissions here.

bastelfreak:
Hi @bitcrush and @c33s, thanks for the work here!

manifests/repo/debian.pp
```diff
@@ -70,22 +74,25 @@
...
     }
 
     ::apt::source { 'source_php_71':
-      location    => 'https://packages.sury.org/php/',
-      release     => $::lsbdistcodename,
-      repos       => 'main',
-      include_src => false,
-      require     => [
+      location => 'https://packages.sury.org/php/',
+      release  => $::lsbdistcodename,
```
bastelfreak: preferred way would be to use the new facts hash instead of a topscope variable. If you want you can change this in another PR, or we just stick with it.

PR #427, opened 2018-03-12
Allow a custom location for the php-fpm error log

DJMuggs:
I like the possibility to choose a custom error log location, very usefull if you use something like RHSCL in combination with this module.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -15,6 +15,7 @@ class php::globals (
   Optional[Pattern[/^[57].[0-9]/]] $php_version = undef,
   Optional[Stdlib::Absolutepath] $config_root   = undef,
   Optional[Stdlib::Absolutepath] $fpm_pid_file  = undef,
+  Optional[String] $fpm_error_log               = undef,
 ) {
 
   $default_php_version = $facts['os']['family'] ? {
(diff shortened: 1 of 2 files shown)


manifests/globals.pp
```diff
@@ -15,6 +15,7 @@
   Optional[Pattern[/^[57].[0-9]/]] $php_version = undef,
   Optional[Stdlib::Absolutepath] $config_root   = undef,
   Optional[Stdlib::Absolutepath] $fpm_pid_file  = undef,
+  Optional[Stdlib::Absolutepath] $fpm_error_log  = undef,
```
c33s: `syslog` must be a valid option here

DJMuggs: syslog is a valid option, see the Archlinux $default_error_log block in the globals.

DJMuggs: Thank you for your reply about the fpm pools, this has helped.

But i still need to handle the different location of the error log file when using an RHSCL php version.
I know there is already a PR for including RHSCL in php module, but that is taking already a while to get merged.

Will fix the coding style.

c33s: i am not sure about the `Stdlib::Absolutepath` will work for `syslog` as far as i can remember, the absolute path must start with a `/` see 
- https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/absolutepath.pp
- https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/unixpath.pp
- https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/windowspath.pp

if i set `php::globals::fpm_error_log: syslog` in my `common.yaml` in hiera, i will get an error. so it should be `Optional` or `Optional[Enum['syslog'], Stdlib::Absolutepath]` or something like that (the correct value has to be tested).

DJMuggs: will take a look at this

c33s: as you reverted the pool change, please rename this PR `removed default www fpm pool and added parameter for error log file` -> `added parameter for error log file`

bastelfreak:
Hi @DJMuggs, thanks for the PR!
Can you squash the commits down a bit? 8 seems to be quite a few for such a change. Can you check the used email address in the commits? It isn't associated with your github account. Also take a look at the spec test, a few are failing.

DJMuggs:
Will squash the commits pretty soon

DJMuggs:
Your test fail because of the manage_repos => true for step 9 in your build ci process.
When you set manage_repos => false , the test for step 9 succeeds, but step 6,7,8 fail.

It has to do with a libdb5.1 that can't be found when using wheezy repos on  jessie.

Can you please fix?

After this my pull request should be without errors.

c33s:
@DJMuggs thank you for your debuging information.  i am working on this issue https://github.com/voxpupuli/puppet-php/issues/433 i came to the same result as you but i haven't made a fix yet. waiting for input on how we should resolve this.

attachmentgenie:
any progress on this ticket, i am using the php packages from remi and run into this problem too

c33s:
@attachmentgenie see my comment above

PR #429, opened 2018-03-14
Do not configure extension if ensure is absent

jkroepke:
Required, if I need to remove an extension.

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -107,7 +107,7 @@ define php::extension (
   }
 
   # PEAR packages don't require any further configuration, they just need to "be there".
-  if $provider != 'pear' {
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @jkroepke, thanks for this. Can you add a spec test that confirms this is now working (or an acceptance test)? There is currently one travis job failing, but that might be related to the debian 8 docker image we use.

jkroepke:
Sure! I also add a Dockerfile a run spec tests locally. Its from https://github.com/voxpupuli/puppet-gitlab/pull/119

It would be a dream to sync this file via modulesync, because I want to setup a local dev environment. Docker is good enough.

jkroepke:
Cool the bot works.

PR #430, opened 2018-03-14
Do not use a a hardcoded default for php::repo::debian

jkroepke:
On Debian 8 and 9 systems, the dotdeb debian 7 repo will be used by default. This shouldn't be.

Some breaking changes here since all location special handling removed for now.

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -17,85 +17,25 @@
 # [*key*]
 #   Public key in apt::key format
 #
-# [*dotdeb*]
-#   Enable special dotdeb handling
-#
-# [*sury*]
-#   Enable special sury handling
-#
 class php::repo::debian(
-  $location     = 'http://packages.dotdeb.org',
-  $release      = 'wheezy-php56',
-  $repos        = 'all',
-  $include_src  = false,
-  $key          = {
-    'id'     => '6572BBEF1B5FF28B28B706837E3F070089DF5277',
-    'source' => 'http://www.dotdeb.org/dotdeb.gpg',
-  },
-  $dotdeb       = true,
-  $sury         = true,
+  String           $location     = 'https://packages.sury.org/php',
+  Optional[String] $release      = undef,
+  String           $repos        = 'main',
+  Boolean          $include_src  = false,
+  Hash             $key          = {
+    'id'     => 'DF3D585DB8F0EB658690A554AC0E47584A7A714D',
+    'source' => 'https://packages.sury.org/php/apt.gpg',
+  }
 ) {
-
   if $caller_module_name != $module_name {
     warning('php::repo::debian is private')
   }
 
-  include '::apt'
-
-  create_resources(::apt::key, { 'php::repo::debian' => {
-    id     => $key['id'],
-    source => $key['source'],
-  }})
-
-  ::apt::source { "source_php_${release}":
+  apt::source { 'php':
     location => $location,
     release  => $release,
     repos    => $repos,
-    include  => {
-      'src' => $include_src,
-      'deb' => true,
-    },
-    require  => Apt::Key['php::repo::debian'],
-  }
-
-  if ($dotdeb) {
-    # both repositories are required to work correctly
-    # See: http://www.dotdeb.org/instructions/
-    if $release == 'wheezy-php56' {
-      ::apt::source { 'dotdeb-wheezy':
-        location => $location,
-        release  => 'wheezy',
-        repos    => $repos,
-        include  => {
-          'src' => $include_src,
-          'deb' => true,
-        },
-      }
-    }
-  }
-
-  if ($sury and $php::globals::php_version == '7.1') {
-    # Required packages for PHP 7.1 repository
-    ensure_packages(['lsb-release', 'ca-certificates'], {'ensure' => 'present'})
-
-    # Add PHP 7.1 key + repository
-    apt::key { 'php::repo::debian-php71':
-      id     => 'DF3D585DB8F0EB658690A554AC0E47584A7A714D',
-      source => 'https://packages.sury.org/php/apt.gpg',
-    }
-
-    ::apt::source { 'source_php_71':
-      location => 'https://packages.sury.org/php/',
-      release  => $facts['os']['distro']['codename'],
-      repos    => 'main',
-      include  => {
-        'src' => $include_src,
-        'deb' => true,
-      },
-      require  => [
-        Apt::Key['php::repo::debian-php71'],
-        Package['apt-transport-https', 'lsb-release', 'ca-certificates']
-      ],
-    }
+    include  => { 'src' => $include_src },
+    key      => $key,
   }
 }
--- a/spec/classes/php_repo_debian_spec.rb
+++ b/spec/classes/php_repo_debian_spec.rb
@@ -1,0 +1,33 @@
+require 'spec_helper'
+
+describe 'php::repo::debian', type: :class do
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let :facts do
+        facts
+      end
+
+      case facts[:osfamily]
+      when 'debian'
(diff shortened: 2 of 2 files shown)


manifests/repo/debian.pp
```diff
@@ -62,8 +62,8 @@
     # both repositories are required to work correctly
     # See: http://www.dotdeb.org/instructions/
     if $release == 'wheezy-php56' {
-      ::apt::source { 'dotdeb-wheezy':
-        location => $location,
+      ::apt::source { 'php-wheezy':
+        location => 'http://packages.dotdeb.org',
```
bastelfreak: we can use $location here, right?

bastelfreak:
Hi @jkroepke, thanks for this bugfix. Can you add unit tests or an acceptance test that verifies it's now working correctly?

jkroepke:
$location is used here

https://github.com/voxpupuli/puppet-php/pull/430/files#diff-9f5ae5c7a7b8c3af838615f1765cde65L51

bastelfreak:
All those repos and conditions are quite confusing. Can you please add tests for this?

manifests/repo/debian.pp
```diff
@@ -24,16 +24,16 @@
...
+  String           $repos        = 'all',
+  Boolean          $include_src  = false,
+  Hash             $key          = {
     'id'     => '6572BBEF1B5FF28B28B706837E3F070089DF5277',
     'source' => 'http://www.dotdeb.org/dotdeb.gpg',
   },
-  $dotdeb       = true,
-  $sury         = true,
+  Boolean          $dotdeb       = false,
+  Boolean          $sury         = false,
```
c33s: Bc break. both bool flags should stay on true or if the `sury` parameter has not be released, we can change sury to false as default

manifests/repo/debian.pp
```diff
@@ -47,7 +47,7 @@
     source => $key['source'],
   }})
 
-  ::apt::source { "source_php_${release}":
+  ::apt::source { 'php-upstream':
```
c33s: @bastelfreak we should make this name configureable. i personally prefer to name my repo files according the upstream name `dotdeb` or `sury` and not according the content `php`.  both provide php but also other things. especially dotdeb provides a lot of other stuff

manifests/repo/debian.pp
c33s: this is the question, if we want this module to auto manage the repos, we have to add hardcoded locations. like 
- if `stretch` and `php version` = `7.1` or `5.6` auto use sury
- if `wheezy` ... use dotdeb or sury (or sury for all, we have to build a matrix what repo should automatically be used

some kind of if location is undef -> use auto mode, else do not use any hardcoded repo

manifests/repo/debian.pp
```diff
@@ -24,16 +24,16 @@
 #   Enable special sury handling
 #
 class php::repo::debian(
-  $location     = 'http://packages.dotdeb.org',
-  $release      = 'wheezy-php56',
-  $repos        = 'all',
-  $include_src  = false,
-  $key          = {
+  String           $location     = 'http://packages.dotdeb.org',
```
c33s: if we use the auto stuff i described below, we also should set the default location and key hash to undef

manifests/repo/debian.pp
jkroepke: I agree, but thats out of the scope here.

jkroepke:
Good enough for most cases.

I removed some special handing. This class should be easier to maintain now.

manifests/repo/debian.pp
```diff
@@ -17,85 +17,25 @@
...
 
-  include '::apt'
-
-  create_resources(::apt::key, { 'php::repo::debian' => {
-    id     => $key['id'],
-    source => $key['source'],
-  }})
-
-  ::apt::source { "source_php_${release}":
+  apt::source { 'php':
```
c33s: should be named `sury` and should be configurable

manifests/repo/debian.pp
```diff
@@ -17,85 +17,25 @@
...
-  $location     = 'http://packages.dotdeb.org',
-  $release      = 'wheezy-php56',
-  $repos        = 'all',
-  $include_src  = false,
-  $key          = {
-    'id'     => '6572BBEF1B5FF28B28B706837E3F070089DF5277',
-    'source' => 'http://www.dotdeb.org/dotdeb.gpg',
-  },
-  $dotdeb       = true,
-  $sury         = true,
```
c33s: we should introduce a enum parameter like `repo_management` `auto|native|sury|(dotdeb)` so the user can decide how the repo management behaves. it should be possible to set a custom repo and prevent the auto added repo from below

manifests/repo/debian.pp
```diff
@@ -17,85 +17,25 @@
...
-        repos    => $repos,
-        include  => {
-          'src' => $include_src,
-          'deb' => true,
-        },
-      }
-    }
-  }
-
-  if ($sury and $php::globals::php_version == '7.1') {
```
c33s: i would prefer keeping the `if` with the `repo_management` parameter. we also could keep the dotdeb repo data in here or better move the hard coded classes to module-hiera-data

manifests/repo/debian.pp
```diff
@@ -17,85 +17,25 @@
...
-      location => 'https://packages.sury.org/php/',
-      release  => $facts['os']['distro']['codename'],
-      repos    => 'main',
-      include  => {
-        'src' => $include_src,
-        'deb' => true,
-      },
-      require  => [
-        Apt::Key['php::repo::debian-php71'],
-        Package['apt-transport-https', 'lsb-release', 'ca-certificates']
```
c33s: why can we drop this?

manifests/repo/debian.pp
```diff
@@ -17,85 +17,25 @@
...
-          'src' => $include_src,
-          'deb' => true,
-        },
-      }
-    }
-  }
-
-  if ($sury and $php::globals::php_version == '7.1') {
-    # Required packages for PHP 7.1 repository
-    ensure_packages(['lsb-release', 'ca-certificates'], {'ensure' => 'present'})
```
c33s: why can we drop this?

c33s:
@jkroepke can we chat somewhere? i already started with the refactoring plans before. mainly i would need help with the broken tests.

manifests/repo/debian.pp
jkroepke: I don't think so. The sury location is just daemon and can be replaced by any other location. I look around other voxpopuli modules. nginx just use nginx, collectd used ppa_collect. Use the software as `apt::source` title should be good enough in my eyes like other modules do it.

manifests/repo/debian.pp
jkroepke: Parameters like `location`, `key` are configurable via hiera. With this PR all standard deb repo like sury are covered by this change. I do not see a point for sury special handing.

manifests/repo/debian.pp
jkroepke: See above.

manifests/repo/debian.pp
jkroepke: apt-transport-https is handed by apt well. 

https://github.com/puppetlabs/puppetlabs-apt/blob/master/manifests/source.pp#L41

`lsb-release` and `ca-certificates` are core packages, its out of scope here. For example apt recommends ca-certificates, you don't need define it here.

jkroepke:
Sure. Does voxpopuli already discord? Or just IRC?

ekohl:
What's the status of this now (other than the merge conflict)?

jkroepke:
@ekohl As I know @c33s doesn't want it, because I drop all the special repo handing which isn't necessary from my side. He wants more puppet classes like `php::repo::debian::sury` but which is in conflict with this PR. Sorry if i'm telling bullshit. Long time ago.

PR #431, opened 2018-03-17
switch the dotdeb repo url to https

bastelfreak:

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -24,7 +24,7 @@
 #   Enable special sury handling
 #
 class php::repo::debian(
-  $location     = 'http://packages.dotdeb.org',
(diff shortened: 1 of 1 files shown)


c33s:
careful with this one. then we need to ensure we also require https://packages.debian.org/stretch/apt-transport-https and maybe also https://packages.debian.org/stretch/ca-certificates
-> bc break and possible trouble/conflicts in existing setups. people partly have this already installed.

maybe we should plan a major release, where we can add #409

bastelfreak:
I need to doublecheck, but I think the puppetlabs-apt module automatically installs the apt-transport-https package if a repo url starts with https

juniorsysadmin:
https://github.com/puppetlabs/puppetlabs-apt/blob/master/manifests/source.pp#L32-L41

PR #432, opened 2018-03-18
WIP:Debian Mysql Extension Fix

c33s:
based upon https://github.com/voxpupuli/puppet-php/pull/322

### DIFF
--- a/manifests/extension/config.pp
+++ b/manifests/extension/config.pp
@@ -83,10 +83,15 @@ define php::extension::config (
   if $provider == 'pear' {
     $final_settings = $full_settings
   } else {
-    $final_settings = deep_merge(
-      {"${extension_key}" => "${module_path}${so_name}.so"},
-      $full_settings
-    )
+    $excluded_add_extension_key_extensions = ['mysql'];
+    if ($so_name in $excluded_add_extension_key_extensions) {
+      $final_settings = $full_settings
+    } else {
+      $final_settings = deep_merge(
+        {"${extension_key}" => "${module_path}${so_name}.so"},
+        $full_settings
+      )
+    }
   }
 
   $config_root_ini = pick_default($::php::config_root_ini, $::php::params::config_root_ini)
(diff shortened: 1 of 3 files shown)


manifests/extension/config.pp
```diff
@@ -102,19 +107,48 @@
...
-      Package[$::php::fpm::package] ~> Exec[$cmd]
+    # run cmds. normally only one command is used for one extension, but debian split up the mysql extension, which now
+    # needs more commands to run.
+    $cmds.each |String $my_so_name, String $my_cmd| {
+      exec { $my_cmd:
+        onlyif  => "${ext_tool_query} -s ${_sapi} -m ${my_so_name} | /bin/grep 'No module matches ${my_so_name}'",
+        require => ::Php::Config[$title],
+      }
+
+      if $::php::fpm {
```
bastelfreak: can you remove the trailing ::, while you are already refactoring? They are totally unneeded.

bastelfreak:
Can you add a few tests or review the current ones so we can ensure that the new logic is correct?

manifests/extension/config.pp
c33s: i would suggest to do this (i can do it) in a seperate PR, i personally dislike having noisy diffs from style fixes together with logical changes.

c33s:
trying to get the tests running locally. 
```
fatal: cannot create directory at 'spec/vcr/GitHubChangelogGenerator_OctoFetcher/_fetch_closed_issues_and_pr/when_API_ca
ll_is_valid/returns_issue_with_proper_key': Filename too long
```

bastelfreak:
The error and the paths look strange. Is this caused by this repo? Which command did you use?

rgevaert:
I'm looking forward to see this completed. For now I see that some other PRs are also being worked on. I'm happy to test some things once things are getting stable.

yakatz:
This PR has had no activity in a long time. Feel free to revise/update/resubmit.

Issue #434, opened 2018-04-04
Wrong merge behavior for settings, extensions, fpm_pools, fpm_global_pool_settings


c33s:
currently the class parameters `settings`, `extensions`, `fpm_pools`, `fpm_global_pool_settings` are getting "double" looked up and merged. 

because of puppets **Automatic class parameter lookup**, which already supports the lookup behaviors `first`, `unique`, `hash` or `deep`, it comes to an unexpected behavior if the `lookup_option` in the hiera file is anything else than `deep` (`first` is the default)

https://puppet.com/docs/puppet/5.3/hiera_automatic.html
https://puppet.com/docs/puppet/5.0/hiera_merging.html

common.yaml
```
lookup_options:
    php::extensions: { merge: {strategy: first} }

php::extensions:
  bcmath: {}
```

my.example.com.yaml
```
php::extensions:
  soap: {}
```

should result in `{ soap }` on `my.example.com` but because of this bug, it results in `{ soap, bcmath }`

the following code snippet where `deep` is hardcoded, is responsible for this bug.
init.pp
```
  # Deep merge global php settings
  $real_settings = deep_merge($settings, lookup('php::settings', {value_type => Hash, merge => 'deep', default_value => {}}))

  # Deep merge global php extensions
  $real_extensions = deep_merge($extensions, lookup('php::extensions', {value_type => Hash, merge => 'deep', default_value => {}}))

  # Deep merge fpm_pools
  $real_fpm_pools = deep_merge($fpm_pools, lookup('php::fpm_pools', {value_type => Hash, merge => 'deep', default_value => {}}))

  # Deep merge fpm_global_pool_settings
  $real_fpm_global_pool_settings = deep_merge($fpm_global_pool_settings, lookup('php::fpm_global_pool_settings', {value_type => Hash, merge => 'deep', default_value => {}}))
```

c33s:
fixed via #435

PR #435, opened 2018-04-04
fixed merge bug

c33s:
see #434

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -149,17 +149,10 @@ class php (
 
   $real_fpm_package = pick($fpm_package, "${package_prefix}${::php::params::fpm_package_suffix}")
 
-  # Deep merge global php settings
-  $real_settings = deep_merge($settings, lookup('php::settings', {value_type => Hash, merge => 'deep', default_value => {}}))
-
-  # Deep merge global php extensions
-  $real_extensions = deep_merge($extensions, lookup('php::extensions', {value_type => Hash, merge => 'deep', default_value => {}}))
-
-  # Deep merge fpm_pools
-  $real_fpm_pools = deep_merge($fpm_pools, lookup('php::fpm_pools', {value_type => Hash, merge => 'deep', default_value => {}}))
-
-  # Deep merge fpm_global_pool_settings
-  $real_fpm_global_pool_settings = deep_merge($fpm_global_pool_settings, lookup('php::fpm_global_pool_settings', {value_type => Hash, merge => 'deep', default_value => {}}))
+  $real_settings = $settings
+  $real_extensions = $extensions
+  $real_fpm_pools = $fpm_pools
+  $real_fpm_global_pool_settings = $fpm_global_pool_settings
 
   if $manage_repos {
     class { 'php::repo': }


bastelfreak:
The next release is probably a major one, we could already drop the old behavior now without all the deprecation boilerplate. I'm a bit worried that we will forget to remove it in the future if we don't do it now. That happend multiple times in the past.

c33s:
@bastelfreak rebased on master and removed deprecation notices (i am clearly against removing the bc layer).

bastelfreak:
Hi @voxpupuli/collaborators, I would like to get some feedback on this. We normally don't add a layer of warnings for future breaking changes. My opinion:

* this adds extra complexity that can introduce bugs
* it's very noisy
* We tend to forget things


I don't want to introduce breaking changes in a minor release. Also they always need to be documented clearly in the CHANGELOG.md and README.md. I think it's fine to have a clearly documented breaking change in the CHANGELOG.md if it's a major release and it doesn't happen on each release. That should not require a deprecation layer.

c33s:
@bastelfreak maybe move this discussion it its own ticket https://github.com/voxpupuli/puppet-php/issues/450 and merge this for now.

xaque208:
I'm not clear why we rename the variables here instead of using the class parameters as intended.  Perhaps the other issues have the context.

c33s:
@xaque208 it is not about this PR please go to #450 in this PR i removed the deprecation warnings. the variable names are a complete different thing.

so at all who come to this ticket because of the @ voxpupuli/collaborators (don't want to notify again) please head to #450,

the code before the discussion of deprecation notices and BC breaks came up looked like this:

init.pp
```
...snip...
  Boolean $force_old_merge_behavior               = true,
) inherits ::php::params {

  $real_fpm_package = pick($fpm_package, "${package_prefix}${::php::params::fpm_package_suffix}")

  if ($force_old_merge_behavior) {
    deprecation('php_hiera_manual_lookup', 'the default deep merge behavior is a bug and will be removed in the next major release')

    # Deep merge global php settings
    $real_settings = deep_merge($settings, lookup('php::settings', {value_type => Hash, merge => 'deep', default_value => {}}))

    # Deep merge global php extensions
    $real_extensions = deep_merge($extensions, lookup('php::extensions', {value_type => Hash, merge => 'deep', default_value => {}}))

    # Deep merge fpm_pools
    $real_fpm_pools = deep_merge($fpm_pools, lookup('php::fpm_pools', {value_type => Hash, merge => 'deep', default_value => {}}))

    # Deep merge fpm_global_pool_settings
    $real_fpm_global_pool_settings = deep_merge($fpm_global_pool_settings, lookup('php::fpm_global_pool_settings', {value_type => Hash, merge => 'deep', default_value => {}}))
  } else {
    $real_settings = $settings

    $real_extensions = $extensions

    $real_fpm_pools = $fpm_pools

    $real_fpm_global_pool_settings = $fpm_global_pool_settings
  }
...snip...
```

c33s:
@bastelfreak any reason why you don't merge it?

bastelfreak:
on stuff that's maybe controversial we want to wait at least 3 days (https://github.com/voxpupuli/plumbing/blob/master/share/governance.md). I want to merge it tomorrow if no other collaborator raises concerns.

c33s:
@bastelfreak is it controversal? i removed the deprecation stuff and only the fix is left. for the controversal stuff i opened up #450

PR #437, opened 2018-04-18
Updated puppet/archive dependency

marknl:
It works fine with puppet/archive 3.0.0, though dep check is < 3.0.0. Currently only able to update with --force

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -27,7 +27,7 @@
     },
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 1.0.0 < 3.0.0"
+      "version_requirement": ">= 1.0.0 <= 3.0.0"
     }
   ],
   "requirements": [


metadata.json
```diff
@@ -27,7 +27,7 @@
     },
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 1.0.0 < 3.0.0"
+      "version_requirement": ">= 1.0.0 <= 3.0.0"
```
alexjfisher: Probably best to make this `< 4.0.0` otherwise it'll need updating as soon as 3.0.1 is released.

marknl: Will do!

alexjfisher: a 4.5.1 of puppet/archive???

marknl: Ah no sorry man, I had the wrong tab opened (puppetlabs-apt) and quotes from there. messing this one up. I'll create a new one with < 4.0.0

alexjfisher:
You didn't really need to close this PR.  Just make the change then commit it with the `--amend` option and then force push your branch.  The PR will be updated automatically.

marknl:
Yeah, I figured that out too late.

alexjfisher:
no worries.  Thanks for the new PR (which I've just merged).

PR #436, opened 2018-04-09
WIP: Debian repo handling refactoring

c33s:

PR #440, opened 2018-04-22
Debian 9 support

SimonHoenscheid:

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -23,6 +23,10 @@ class php::globals (
         /^(1[67].04)$/ => '7.0',
         default => '5.x',
       },
+      'Debian' => $facts['os']['release']['major'] ? {
+        '9' => '7.0',
+        default => '5.x',
+      },
       default => '5.x',
     },
     default => '5.x',
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -17,16 +17,14 @@ class php::globals (
   Optional[Stdlib::Absolutepath] $fpm_pid_file  = undef,
 ) {
 
-  $default_php_version = $facts['os']['family'] ? {
-    'Debian' => $facts['os']['name'] ? {
-      'Ubuntu' => $facts['os']['release']['full'] ? {
-        /^(1[67].04)$/ => '7.0',
-        default => '5.x',
-      },
-      'Debian' => $facts['os']['release']['major'] ? {
-        '9' => '7.0',
-        default => '5.x',
-      },
+  $default_php_version = $facts['os']['name'] ? {
+    'Debian' => $facts['os']['release']['major'] ? {
+      '9' => '7.0',
+      default => '5.x',
+    },
+    'Ubuntu' => $facts['os']['release']['major'] ? {
+      '18.04' => '7.2',
+      '16.04' => '7.0',
       default => '5.x',
     },
     default => '5.x',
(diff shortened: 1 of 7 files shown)


bastelfreak:
Thanks for this. However, we need to merge https://github.com/voxpupuli/puppet-php/pull/436 before

bastelfreak:
thanks for the PR @SimonHoenscheid. Can you take a look at the used email address in the commits? It isn't associated with your github account.

bastelfreak:
Can you take a look at the rubocop issues in https://travis-ci.org/voxpupuli/puppet-php/jobs/369870375 ?

SimonHoenscheid:
@bastelfreak old employer email, no chance i add this again, I have a look if I can amend these commits and push them again.
Sure.

bastelfreak:
```
git commit --amend --author="John Doe <john@doe.org>"
``` 

should work

c33s:
if the test are ok i think we can merge this, i can rebase on it as long as only the current files are touched. but we have still the problem with the breaking tests (had no time to debug this).

manifests/globals.pp
```diff
@@ -23,6 +23,10 @@
         /^(1[67].04)$/ => '7.0',
         default => '5.x',
       },
+      'Debian' => $facts['os']['release']['major'] ? {
```
c33s: i would love to see this section refactored a little bit.
`['os']['distro']['id']` which allows a clear seperation between ubuntu and debian.

c33s: > OSes, they are more complex than they need to be, maybe its also a good thing to clean them up later
@SimonHoenscheid do you ment this part with you comment in the issue? 
after a little more thinking this refactoring of this would be better done after merging my repo refactoring because most of the stuff then can stay in the hiera.yaml

SimonHoenscheid: yes exactly. dividing Ubuntu and Debian here completely would be much more clear

spec/classes/php_spec.rb
```diff
@@ -7,27 +7,50 @@
         facts
       end
 
+      php_cli_package = case facts[:osfamily]
+                        when 'Debian'
+                          case facts[:os]['release']['major']
```
c33s: can't we use the correct facts here?
`Debian` with version `16.04` is confusing. would prefer a switch between debian and ubuntu here.

c33s: and thank you for your PR. n1!

SimonHoenscheid: Then I would suggest splitting these two up completely

manifests/globals.pp
SimonHoenscheid: I will have a look at this later.

bastelfreak:
Should we remove Debian 8? The repo seems to be broken. I'm not sure if that's something that we can fix on our end.

SimonHoenscheid:
Which repo? Debian 8 is EOL in June with LTS Support until April 2020.

SimonHoenscheid:
I have some time off end of the week and will clean up the conditions in puppet code and spec then.

SimonHoenscheid:
I updated the conditions as requested but did not use the  facter os.distro hash, because it is missing in older facter versions. I used  facter os.name and facter os.release.major instead which should be available in older versions.

SimonHoenscheid:
added acceptence tests and added Travis CI Job

SimonHoenscheid:
The acceptence test for debian 8 with php 5.6 . In my eyes the test is kind of useless, because php 5.6 is the default package coming with debian 8 and this is already covered by the normal test

bastelfreak:
Hey people, thanks for figuring this out!

PR #455, opened 2018-06-11
allow php 7.2

cbergmann:
Sury supports php 7.2 for debian. As php 7.1 (and 7.0) will be discontinued end of this year it might be needed to migrate to 7.2. As 7.0 is the only supported Version in official Debian 9 repos. I think that this Module should support PHP 7.2 from sury.

n/a


manifests/globals.pp
```diff
@@ -77,6 +77,7 @@
         }
       } else {
         case $globals_php_version {
+          /^5\.6/,
```
igalic: where was this before?

igalic: yupp, that's what travis is showing https://travis-ci.org/voxpupuli/puppet-php/jobs/391234923#L1338

cbergmann: so what should We Do? Drop Debian 7 Support or more code?

manifests/repo/debian.pp
```diff
@@ -72,17 +72,17 @@
...
-    # Add PHP 7.1 key + repository
-    apt::key { 'php::repo::debian-php71':
+    # Add PHP sury key + repository
+    apt::key { 'php::repo::debian-php-sury':
       id     => 'DF3D585DB8F0EB658690A554AC0E47584A7A714D',
       source => 'https://packages.sury.org/php/apt.gpg',
     }
 
-    ::apt::source { 'source_php_71':
+    ::apt::source { 'source_php_sury':
```
bastelfreak: Is that the name of the created file in `/etc/apt/sources.list.d/`? So people would end up with two files that contain the same repo?

manifests/globals.pp
bastelfreak: Everything that is Debian 7 specific can go away. We don't support it anymore. But we still have to support PHP 5.6.

manifests/repo/debian.pp
dsampaolo: I think it's a good idea to remove the "7.1" mark in the repo name. Sury is stable and will be used too for future versions.

manifests/globals.pp
bastelfreak: actually, we could drop PHP 5.6 now.

PR #474, opened 2018-09-05
modulesync 2.0.0

bastelfreak:


bastelfreak:
Do we really want to test against debian 7? mhm...

Issue #477, opened 2018-09-19
PHP modules are enabled even when ensure is absent


amateo:
I have this code
```
class {'php':
  extensions => {
    'mysql' => {
      ensure => 'absent',
    }
  }
}
```

with this code, package for php mysql is removed from the system, but the module is forced to be enabled with:
```
Notice: /Stage[main]/Php/Php::Extension[mysql]/Php::Extension::Config[mysql]/Php::Config[mysql]/Php::Config::Setting[/etc/php/7.0/mods-available/mysql.ini: extension]/Ini_setting[/etc/php/7.0/mods-available/mysql.ini: extension]: Triggered 'refresh' from 1 event
Notice: /Stage[main]/Php/Php::Extension[mysql]/Php::Extension::Config[mysql]/Exec[/usr/sbin/phpenmod -v 7.0 -s ALL mysql]/returns: executed successfully
Notice: /Stage[main]/Php/Php::Extension[mysql]/Php::Extension::Config[mysql]/Exec[/usr/sbin/phpenmod -v 7.0 -s ALL mysql]: Triggered 'refresh' from 1 event
```

amateo:
Just provided in #478.

TuningYourCode:
This issue can be closed as it is resolved?

PR #478, opened 2018-09-25
Remove config when module is ensured to absent

amateo:
When the module is ensured to absent, it removes its config and not try
to create the config file and enable it.


bastelfreak:
Fixes https://github.com/voxpupuli/puppet-php/issues/477

manifests/extension/config.pp
```diff
@@ -90,31 +90,37 @@
...
+        'ALL' => 'cli',
+        default => $sapi,
+      }
+      exec { $cmd:
+        onlyif  => "${ext_tool_query} -s ${_sapi} -m ${so_name} | /bin/grep 'No module matches ${so_name}'",
+        require => ::Php::Config[$title],
+      }
     }
 
     if $php::fpm {
```
bastelfreak: I think this is specific to `Debian` as well?

bastelfreak:
Hi @amateo, thanks for this PR as well! Can you take a look at travis? The jobs are failing on CentOS, probably because of the if block in line 118 (see my inline comment for that). Could you add an acceptance test for this so we know it's working properly now?

amateo:
Hi @bastelfreak. I think all problems are now solved.

RedChops:
This patch ends up being a bit of a wild goose chase for all the available values of $ensure, since it can also be 'snapshot' or a pinned version. Even with #485 these will be marked as 'absent'. Doesn't it make more sense for line 93 to simply be `if $ensure != 'absent'`?

amateo:
I have just made #493 with this last change.

PR #479, opened 2018-09-25
Fix enable extension when there is no module associated

amateo:
When the extension has no module associated (e.g. with net-curl), the
module always run phpenmod command (because phpquery always report that
the module is not matched).
This commit tries to fix this problem

### DIFF
--- a/manifests/extension/config.pp
+++ b/manifests/extension/config.pp
@@ -108,13 +108,15 @@ define php::extension::config (
       'ALL' => 'cli',
       default => $sapi,
     }
-    exec { $cmd:
-      onlyif  => "${ext_tool_query} -s ${_sapi} -m ${so_name} | /bin/grep 'No module matches ${so_name}'",
-      require => ::Php::Config[$title],
-    }
+    if has_key($final_settings, 'extension') and $final_settings[extension] {
+      exec { $cmd:
+        onlyif  => "${ext_tool_query} -s ${_sapi} -m ${so_name} | /bin/grep 'No module matches ${so_name}'",
+        require => ::Php::Config[$title],
+      }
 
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @amateo, thanks for the patch! Could you add an acceptance test to https://github.com/voxpupuli/puppet-php/tree/master/spec/acceptance to verify that this now works? Probably just install `net-curl`?

amateo:
Hi @bastelfreak.
The error of last push I think it is not directly related with my changes, because it fails with:
```
An error occurred while loading ./spec/acceptance/php56_spec.rb.
Failure/Error: require 'beaker-rspec'
Docker::Error::UnexpectedResponseError:
```
Anyway, I'm not really sure the test added to acceptance has any sense. The problem for this PR is that there are some php extensions that don't have any `.so` file. In these cases, the module always tries to execute the `phpenmod` en every puppet run. This PR fixes this problem.

The problem with the acceptance test is that I haven't found any of these extension in centos servers.
I previously added the instalation of `url` extension (which is the one I'm having problem in my systems), but this extension didn't exists in centos, so it fails. Then I changed it for `curl`, but this extension do have a `.so` file, so it is not the real test I wanted.

bastelfreak:
Hey @amateo, you can create a new spec file and only execute it on debian. The testsetup takes a minute or two to read. It is defined in: https://github.com/voxpupuli/puppet-php/blob/master/.travis.yml

basically we execute each spec file for each OS for puppet 5 and puppet 6. You would also need to adjust https://github.com/voxpupuli/puppet-php/blob/master/.sync.yml, which is source for generating the .travis.yml from time to time.

bastelfreak:
thanks!

PR #483, opened 2018-10-09
Allow `clear_env` to be disabled

joshuaspence:
Allow `clear_env` to be disabled for PHP-FPM pool definitions.


manifests/fpm/pool.pp
```diff
@@ -149,6 +152,7 @@
   $include                                 = undef,
   $env                                     = [],
   $env_value                               = {},
+	$clear_env                               = undef,
```
bastelfreak: Can you please fix the indentation?

bastelfreak:
Thanks @joshuaspence !

PR #484, opened 2018-10-10
Remove readable for others on pool config file

l-lotz:
The pool config does not need to be readable by others. This allows putting secrets into the files, without them being exposed. It might make sense to make this configurable.


bastelfreak:
thanks @l-lotz !

PR #485, opened 2018-10-11
ensure could also be 'latest'

amateo:
`ensure` parameter could also have the value `latest`, so this commit consider this
case.


bastelfreak:
Thanks for the fix @amateo !

PR #494, opened 2018-10-31
Rely more on puppetlabs-apt

ekohl:
This refactors the code to rely more on puppetlabs-apt features. It passes in the key and lets apt::source define apt::key for us.

Puppetlabs-apt also knows better when apt-transport-https is actually needed. On the most recent Debian/Ubuntu releases it doesn't even exist anymore since apt handles it out of the box.

It also removes the absolute resource notation which is no longer needed.

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -38,14 +38,9 @@ class php::repo::debian(
 
   assert_private()
 
-  include '::apt'
+  include 'apt'
 
-  create_resources(::apt::key, { 'php::repo::debian' => {
-    id     => $key['id'],
-    source => $key['source'],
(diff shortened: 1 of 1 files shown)


ekohl:
@redchops would you be able to test and verify this works for you?

RedChops:
@ekohl I just implemented the patch and it seems to have applied nicely without a change on my nodes!

ekohl:
Thanks for testing @RedChops!

I've marked it as enhancement because strictly speaking we do change the behavior a bit. By passing $key in directly it's possible to directly change the `apt::key` resource. Also it's not a trivial bug fix IMHO.

PR #498, opened 2018-11-16
Fix extension config for zend_extensions

mpdude:
Fixes a glitch introduced in #479.

When an extension is `zend = true`, there is no `extension` entry in the `$final_settings` hash, but a `zend_extension` one.

/cc @amateo


bastelfreak:
Hi @mpdude, thanks for the patch. Are you able to contribute a unit- or acceptance test?

mpdude:
#567 seems to resolve this.

PR #508, opened 2019-02-21
Adding php 7 support for rhel family

mgrune9978:


mgrune9978:
Fixing issue with top level variable I didn't notice...

PR #513, opened 2019-04-07
modulesync 2.12.0 and drop puppet 4

bastelfreak:
modulesync 2.7.0

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -19,7 +19,4 @@
     - set: debian9-64
       options:
         script: 'bundle exec rspec spec/acceptance/php_spec.rb'
-    - set: debian8-64
-      options:
-        script: 'bundle exec rspec spec/acceptance/php_spec.rb'
   secure: "GOhttACuJt+3s38m4WnW5RuTgwqaAoeEQnNT+X1Ukn7KdcIk4KV8NzYU/CC0VIm8lUOnBWYJKEC4ixX/J/4Wbxox2RAoKMQrO++L0DB1zTCJnq9SfoUBMaQhXvLu+PbxAR0p3P47ozra0C+pOWDpOaxT9ecufrPQt9W9Z4aY/bs="
--- a/spec/acceptance/php_spec.rb
+++ b/spec/acceptance/php_spec.rb
@@ -14,8 +14,6 @@ describe 'php with default settings' do
       packagename = 'php7.2-fpm'
     when %r{ubuntu-16.04}
       packagename = 'php7.0-fpm'
-    when %r{ubuntu-14.04}
-      packagename = 'php5-fpm'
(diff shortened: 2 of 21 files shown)


c33s:
is it really a good step to remove puppet4 support? what are the benefits of dropping the support for puppet4? 

debians current lts is supported until June 2022 (stretch) which ships puppet 4.8. you are defacto killing the native puppet support in currents debian stable.

https://en.wikipedia.org/wiki/Debian_version_history#Debian_9_(Stretch)

bastelfreak:
Hi @c33s. This is the result of a pretty long discussion which happend in https://github.com/voxpupuli/voxpupuli.github.io/pull/157.

dhoppe:
These tests must be converted from mocha to rspec-mocks.

EmilienM:
LGTM

.sync.yml
```diff
@@ -19,7 +19,4 @@
     - set: debian9-64
       options:
         script: 'bundle exec rspec spec/acceptance/php_spec.rb'
-    - set: debian8-64
```
dhoppe, line 22: @bastelfreak We should keep this, since Debian 8 is not EOL until 2020-06-30.

spec/acceptance/php_spec.rb
```diff
@@ -10,12 +10,10 @@
     end
 
     case default[:platform]
-    when %r{ubuntu-18.04}
```
ekohl, line 13: I'm generally a fan of smaller diffs and keeping it newest to oldest is OK IMHO. It's still sorted and consistent.

spec/unit/provider/package/pear_spec.rb
```diff
@@ -33,46 +33,46 @@
...
       expect(described_class.instances).to eq []
     end
   end
 
   describe '#install' do
     it 'installs a package' do
-      described_class.expects(:pear).
+      allow(described_class).to receive(:pear).
         with('-D', 'auto_discover=1', 'upgrade', '--alldeps', 'dummy')
       provider.install
```
ekohl, line 48: To really replace this and keep the test, you should add a `have_received` assertion. This is repeated so I won't comment it everywhere.
```suggestion
      provider.install
      expect(described_class).to have_receive(:pear)
```

dhoppe: I already tried this, but it did not work as expected:
* https://travis-ci.org/github/voxpupuli/puppet-php/jobs/672461881#L1193

bastelfreak:
as this finally works, I'm going to merge it for now. We can still improve it later on.

c33s:
yeah no puppet4 drop :+1:  :smile: :100: :1st_place_medal:

PR #515, opened 2019-04-15
add pool_purge option to php init.pp

AdamJacobMuller:
It currently appears to be very difficult to set pool_purge. I suppose it is possible by setting fpm => false and including classes manually but this seems like a much cleaner option which is in-line with how other settings are done in the package.

### DIFF
--- a/manifests/fpm.pp
+++ b/manifests/fpm.pp
@@ -49,6 +49,10 @@
 #   Hash of defaults params php::fpm::pool resources that will be created.
 #   Defaults is empty hash.
 #
+# [*pool_purge*]
+#   Whether to purge pool config files not created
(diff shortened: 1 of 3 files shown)


manifests/fpm.pp
```diff
@@ -64,6 +68,7 @@
   Hash $pools                   = $php::real_fpm_pools,
   $log_owner                    = $php::log_owner,
   $log_group                    = $php::log_group,
+  $pool_purge                   = $php::pool_purge,
```
natemccurdy: Please add a `Boolean` data type to this parameter.

dhoppe:
This has been solved by #557.

PR #522, opened 2019-05-30
Fix deprecated Hiera lookup warnings. Add default www pool in YAML.

### LINKED ISSUE #449: Replace hiera with lookup
Time goes on, hiera lookups are going to be deprecated, we should replace them with the appropriate lookup functions

comport3:
Fix deprecated Hiera lookup warnings. Add YAML structure for default www pool so it can be overridden as needed.

### DIFF
--- a/data/default.yaml
+++ b/data/default.yaml
@@ -1,0 +1,13 @@
+---
+php::fpm_pools:
+  www:
+    catch_workers_output: 'no'
+    listen: '127.0.0.1:9000'
+    listen_backlog: '-1'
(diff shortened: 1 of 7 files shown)


bastelfreak:
Hi @comport3, thanks for the PR!

c33s:
shouldn't `lookup` functions should be avoided? they can lead to uncontrolled behavior as you can't override hardcoded lookup parameters. https://github.com/voxpupuli/puppet-php/issues/434

PR #525, opened 2019-07-05
Remove require php::dev when Ubuntu 14.04

guimaluf:
When using Ubuntu 14.04 there is no such package as php-dev

This commits remove the need of php::dev for extensions when running on
Ubuntu 14.04

TODO: create spec tests for this scenario


guimaluf:
without the "dev" package there is no way to run the `phpize` in ubuntu 14.04

PR #526, opened 2019-07-13
Simplify php (extension) removal

TuningYourCode:
This PR enables simplified removal of php. Without this change you have to set each extension the ensure attribute itself because php::ensure: absent will currently not influence php::extension::$title::ensure.

### DIFF
--- a/manifests/extension.pp
+++ b/manifests/extension.pp
@@ -4,7 +4,7 @@
 #
 # [*ensure*]
 #   The ensure of the package to install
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @TuningYourCode, thanks for the PR. Can you add some tests to this?

TuningYourCode:
Tests added. Should also provide basic test if packages get removed on $php::ensure = 'absent';

bastelfreak:
Thanks for the PR @TuningYourCode !

Issue #529, opened 2019-08-07
No default version for debian buster in globals


lelutin:
The globals.pp file currently does not define a default version for debian buster (10).

Because of this, when using this module on buster, the module defaults to php 5.x (wth).

A simple fix would be to add a case for the major version 10 of debian to make it install php 7.3

PR #530, opened 2019-08-14
set default version for debian buster and add support for buster

lelutin:
debian buster was released on july 6th and many folks are already
deploying it. This module should start supporting the new debian stable
version.

For this, we need the module to know what the new default PHP version is
for this new release.

Closes: #529


lelutin:
my last comment fixed the checks but travis is buggy and a bunch of checks timed out (hence the errors that show up). I don't know how to relauch the jobs that errored out

lelutin:
This time most of the workers completed their jobs but two of them got networking issues >.<

PR #531, opened 2019-08-17
Prevent creation of pid file directory if not used

### LINKED ISSUE #501: undesired service restart due to missing /var/run/php-fpm dir

- Puppet: 6.1.0
- Ruby: 2.5.3
- Distribution: Ubuntu Bionic
- Module version: 6.0.2

Create a FPM pool that matches Ubuntu default config as much as possible:

```
    case $::os['release']['major'] {
      '18.04': { $php_version = '7.2' }
      '16.04': { $php_version = '7.0' }
      '14.04': { $php_version = '5' }
      default: { fail("${module_name} does not support this OS version") }
    }
    class { 'php::globals':
      fpm_pid_file => "/run/php/php${php_version}-fpm.pid",
    }
    class { 'php':
      dev                      => false,
      composer                 => false,
      pear                     => false,
      fpm_global_pool_settings => {
        'listen'               => "/run/php/php${php_version}-fpm.sock",
        'listen_backlog'       => '511',
        'listen_owner'         => 'www-data',
        'listen_group'         => 'www-data',
        'pm_max_children'      => '5',
        'pm_start_servers'     => '2',
        'pm_min_spare_servers' => '1',
        'pm_max_spare_servers' => '3',
        'slowlog'              => 'log/$pool.log.slow',
      },
    }
```

## What are you seeing

The unneeded directory **/var/run/php-fpm** gets created during the first puppet agent run after boot and triggers an undesired service restart:

```
# puppet agent -t  --noop
Info: Using configured environment 'production'
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Retrieving locales
Info: Loading facts
Info: Applying configuration version '1545336079'
Notice: /Stage[main]/Php::Fpm::Config/File[/var/run/php-fpm/]/ensure: current_value 'absent', should be 'directory' (noop)
Notice: Class[Php::Fpm::Config]: Would have triggered 'refresh' from 1 event
Info: Class[Php::Fpm::Config]: Scheduling refresh of Class[Php::Fpm::Service]
Notice: Class[Php::Fpm::Service]: Would have triggered 'refresh' from 1 event
Info: Class[Php::Fpm::Service]: Scheduling refresh of Service[php7.2-fpm]
Notice: /Stage[main]/Php::Fpm::Service/Service[php7.2-fpm]: Would have triggered 'refresh' from 1 event
Notice: Class[Php::Fpm::Service]: Would have triggered 'refresh' from 1 event
Notice: Class[Php::Fpm]: Would have triggered 'refresh' from 2 events
Notice: Class[Php]: Would have triggered 'refresh' from 1 event
Notice: Stage[main]: Would have triggered 'refresh' from 4 events
Notice: Applied catalog in 3.24 seconds
```

The useless directory **/var/run/php-fpm** shouldn't be created and no service restart should happen due to it.

TuningYourCode:
This PR prevents the creation of the hardcoded pid file directory if it is not used.

Fixes #501


manifests/fpm/config.pp
```diff
@@ -111,7 +111,15 @@
     mode    => '0644',
   }
 
-  ensure_resource('file', ['/var/run/php-fpm/', '/var/log/php-fpm/'], {
+  if (dirname($pid_file) == '/var/run/php-fpm') {
+    ensure_resource('file', ['/var/run/php-fpm/'], {
```
bastelfreak: Would it make sense to always create the `dirname($pid_file)` directory, and not only if it's `/var/run/php-fpm`?

TuningYourCode: Well i see multiply issues if we always do that:
If $pid_file is something like /var/run/php-fpm.pid we would manage /var/run which is most likly something we don‘t want to do.

If $pid_file something complete custom (like /var/run/custom/location/php-fpm.pid) it would be kind of weird that we manage a part of it.

Best solution would be probably get rid of this totally and rely on the package management of the distribution to create the folders and adjust our defaults to represent the default configuration of the distribution.

bastelfreak: Honestly, I was surprised to discover the file resource in the code. I also had the assumption that systemd would create the directories during the service startup. I'm not sure what is the best way to proceed here.

TuningYourCode: I removed it now entirly so that we rely on packet managment to create the default folders.
People which use custom locations should anyway care about the creation themself.

Tested already debian stretch and buster, which work perfectly with this change.

TuningYourCode:
@bastelfreak any feedback on this PR or should we just close and leave the code as it is?

manifests/fpm/config.pp
```diff
@@ -111,12 +111,6 @@
     mode    => '0644',
   }
 
-  ensure_resource('file', ['/var/run/php-fpm/', '/var/log/php-fpm/'], {
```
ghoneycutt: The issue reports a problem with `/var/run/php-fpm/` not `/var/log/php-fpm/`.

TuningYourCode: Well we could of course keep the `/var/log/php-fpm/` file resource. But if we let manage the run directory by the package management why shouldn't we rely on the package management to create the log directory too?
It's also possible to configure the pool to not use the directory (and it is hardcoded here and not dependent on other configurations).

TuningYourCode:
Hi @bastelfreak & @ghoneycutt,

how should we proceed with this PR? If there are any todos please tell me :)
Feel free to close this PR if you do not like to merge it.

manifests/fpm/config.pp
ghoneycutt: The current code allows for specify a different user/group from what the package management uses. By removing the log file directory you remove functionality from the module. 

Also the log file directory is not related to the issue with the pid file. By splitting up different functionality into different PR's, it makes it easier to merge.

ghoneycutt:
Please change to only remove the pid file directory and I'm +1 for merging.

TuningYourCode:
Restored the logic to manage the log directory :)

bastelfreak:
@TuningYourCode there are merge conflicts and they must be resolved before we can review and merge this.

TuningYourCode:
Not sure why these tests fail, they do not check for presents of the removed definition.

PR #533, opened 2019-09-11
fix "cannot redefine $real_settings" error

crispygoth:
Trying to use the apache_config class to apply settings to the apache
SAPI fails with an error caused by a duplicate variable definition:

```
Error while evaluating a Resource Statement, Evaluation Error: Cannot reassign variable '$real_settings' (file:
/etc/puppetlabs/code/environments/development/modules/php/manifests/apache_config.pp, line: 18, column: 18)
```

### DIFF
--- a/manifests/apache_config.pp
+++ b/manifests/apache_config.pp
@@ -15,7 +15,7 @@ class php::apache_config(
 
   assert_private()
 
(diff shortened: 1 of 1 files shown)


natemccurdy:
This looks sane to me. Definitely a syntax error caused by the previous change to this file in #522 

This manifest could use some tests though to prevent this from happening again.

PR #550, opened 2019-10-18
Do a `deep` merge on `fpm` lookup

### LINKED ISSUE #536: Fpm config results in "expects a Hash value, got Tuple"
- Puppet: 5.5
- Ruby: 2.4
- Distribution: Debian 10 (buster)
- Module version: 7.0.0

In the manifest:
```puppet
  include ::php
```
In the hiera config:
```hiera
php::fpm::pools:
  www:
    listen: '/var/run/php-fpm.sock'
    listen_owner: 'www-data'
    listen_group: 'www-data'
    pm_max_children: '100'
    pm_status_path: '/fpmstatus'
```

## What are you seeing
Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a Function Call, Found value has wrong type, expects a Hash value, got Tuple (file: /etc/puppetlabs/code/environments/php
update/modules/php/manifests/fpm.pp, line: 103, column: 17)

The pool config should have been pushed to the node. Whatever I change to the hiera config, it always ends with the same error.

## Output log
See above

## Any additional information you'd like to impart
This behaviour is seen since the new module version is released. With the old version there are no issues.

sigv:
When allowing configuring `php::fpm::pools` as hashes in Hiera, the `unique` ("array merge") behavior is currently used. [Documentation](https://puppet.com/docs/puppet/6.4/function.html#merge-behaviors) states this will return a merged array even though we are interested in a hash. Instead of doing this manual lookup, we can switch to auto lookups.

Fixes #536

### DIFF
--- a/data/default.yaml
+++ b/data/default.yaml
@@ -2,7 +2,7 @@
 
 lookup_options:
   php::fpm::pools:
-    merge: hash
+    merge: first
 
 php::fpm::pools:
   www:
--- a/data/default.yaml
+++ b/data/default.yaml
@@ -1,4 +1,9 @@
 ---
+
+lookup_options:
+  php::fpm::pools:
+    merge: hash
+
 php::fpm::pools:
   www:
     catch_workers_output: 'no'
--- a/data/default.yaml
+++ b/data/default.yaml
@@ -1,5 +1,5 @@
 ---
-php::fpm_pools:
+php::fpm::pools:
   www:
     catch_workers_output: 'no'
     listen: '127.0.0.1:9000'
--- a/README.md
+++ b/README.md
@@ -240,6 +240,23 @@ php::fpm::pools:
 
 ## Notes
 
+### Inheriting configuration across mutliple Hiera sources
+
+Configuration from Hiera such as `php::fpm::pools` is automatically
+lookup up using the "first" merge method. This means that the first
+value found is used. If you instead want to merge the hash keys
+across multiple Hiera sources, you can use [`lookup_options`] to set
+[`hash` or `deep` behaviors] such as:
+
+```yaml
+lookup_options:
+  php::fpm::pools:
+    merge: hash
+```
+
+[`lookup_options`]: https://puppet.com/docs/puppet/6.4/hiera_merging.html#concept-2997
+[`hash` or `deep` behaviors]: https://puppet.com/docs/puppet/6.4/hiera_merging.html#merge-behaviors
+
 ### Debian squeeze & Ubuntu precise come with PHP 5.3
 
 On Debian-based systems, we use `php5enmod` to enable extension-specific
--- a/README.md
+++ b/README.md
@@ -245,8 +245,9 @@ php::fpm::pools:
 Configuration from Hiera such as `php::fpm::pools` is automatically
 lookup up using the "first" merge method. This means that the first
 value found is used. If you instead want to merge the hash keys
-across multiple Hiera sources, you can use [`lookup_options`] to set
-[`hash` or `deep` behaviors] such as:
+across multiple Hiera sources, you can use [`lookup_options`] to
+set [`hash` or `deep` behaviors] such as in the example
+[data/default.yaml](data/default.yaml):
 
 ```yaml
 lookup_options:
--- a/manifests/fpm.pp
+++ b/manifests/fpm.pp
@@ -70,7 +70,7 @@ class php::fpm (
     warning('php::fpm is private')
   }
 
-  $real_settings = lookup('php::fpm::settings', Hash, {'strategy' => 'deep', 'merge_hash_arrays' => true}, $settings)
+  $real_settings = $settings
 
   # On FreeBSD fpm is not a separate package, but included in the 'php' package.
   # Implies that the option SET+=FPM was set when building the port.
@@ -99,8 +99,8 @@ class php::fpm (
 
   Class['php::fpm::config'] ~> Class['php::fpm::service']
 
-  $real_global_pool_settings = lookup('php::fpm::global_pool_settings', Hash, {'strategy' => 'unique'}, $global_pool_settings)
-  $real_pools = lookup('php::fpm::pools', Hash, {'strategy' => 'unique'}, $pools)
+  $real_global_pool_settings = $global_pool_settings
+  $real_pools = $pools
   create_resources(::php::fpm::pool, $real_pools, $real_global_pool_settings)
 
   # Create an override to use a reload signal as trusty and utopic's


bastelfreak:
Hi @sigv, thanks for the fix. Can you add a test for it?

sigv:
The `real_fpm_pools` variable from `init` class is assigned as `pools` variable on `fpm` class. That is a fallback for `real_pools` variable in `fpm` class. That is:

```
class php (
  Hash $fpm_pools                                 = {},
) {
  $real_fpm_pools = $fpm_pools
}
```

```
class php::fpm (
) {
  $real_pools = lookup('php::fpm::pools', Hash, {'strategy' => 'deep'}, $pools)
  create_resources(::php::fpm::pool, $real_pools, $real_global_pool_settings)
}
```

Documentation suggests setting `php::fpm::pools` directly from Hiera, which then overrides the `init` fallback using `lookup`. My proposal then is to test `default.yaml` with a configuration based on the documentation.

c33s:
@bastelfreak please **don't** merge, because we should go in the oposite direction and get rid of ALL hardcoded lookups. the hardcoded lookups break puppets ability to define the merge behavior in hiera files (Automatic class parameter lookup).

@sigv you can define the merge behavior you want in your hiera. in your case simply add this to your hiera file (you can add it to your global hiera file and you can also change the lookup behavior from your global hiera file on any other hiera level.

```
lookup_options:
    php::fpm::pools: { merge: {strategy: deep} }
```

also see my issue https://github.com/voxpupuli/puppet-php/issues/434

sigv:
@c33s I do agree that variables should be pulled in auto-magically without using the `lookup` function, however the current merge strategy does not work outright. [Documentation](https://puppet.com/docs/puppet/6.4/function.html#merge-behaviors) explicitly states `unique` combines arrays and scalar values returning a merged, flattened array with duplicate values removed, meanwhile `deep` combines keys and values of hashes returning a merged hash. I do not think it makes sense for me to rework the two individual variables that are affected here into automatically being looked up without the use of the explicit function as the whole project utilizes that methodology anyhow and am instead proposing patching the merge behavior. This does not in of itself block an upcoming rewrite, but does fix a current bug.

sigv:
I could switch it out to be a `first` lookup instead if that makes more sense as a default.

bastelfreak:
My personal opinion is to get rid of all those lookup calls and use automatic class parameter lookup. That's also the prefered way in the puppet style guide.

sigv:
Is the default merge behavior change a no-go for `deep` and no-go for `first`?

sigv:
The current state of PR should have the same behavior (`first`) that automatic lookup has. That should mean a smooth migration down the line. I do agree that making it automatic lookup is nicer, but is that really a reason for holding a small bugfix patch back?

Dan33l:
@sigv i  don't understand why defining merging behavior in your hiera data doesn't it fit your needs?

You can try this syntax also:
```
lookup_options:
  php::fpm::pools:
    merge: deep
```

sigv:
Of course I could add this and I get the default auto lookup behavior as well:

```
lookup_options:
  php::fpm::pools:
    merge: first
```

At the current state, if someone follows documentation on how to configure via Hiera, they add a `php::fpm::pools` Hash in their Hiera. That does not work on its own as of right now without forcing `lookup_options` to be something else. This PR fixes that problem.

If the `php::fpm::pools` is not expected to work on its own, documentation needs be updated to reflect the requirement on `lookup_options`.

c33s:
@sigv

> however the current merge strategy does not work outright. [Documentation](https://puppet.com/docs/puppet/6.4/function.html#merge-behaviors) explicitly states `unique` combines arrays and scalar values returning a merged, flattened array with duplicate values removed, meanwhile `deep` combines keys and values of hashes returning a merged hash.

this is a good catch, thank you very much for that.

> I do not think it makes sense for me to rework the two individual variables that are affected here into automatically being looked up without the use of the explicit function as the whole project utilizes that methodology anyhow and am instead proposing patching the merge behavior. 

my opinion is different here, if we touch something, it should be better afterwards. for me the argument does not count to say "it makes no sense to clean a room because the whole flat is dirty".

as far as i can see its just:


https://github.com/voxpupuli/puppet-php/blob/d711fc78789e189ace974e998fc72c80aec82dc2/manifests/fpm.pp#L102
```
$real_global_pool_settings = lookup('php::fpm::global_pool_settings', Hash, {'strategy' => 'first'}, $global_pool_settings)
$real_pools = lookup('php::fpm::pools', Hash, {'strategy' => 'first'}, $pools)
```

->

```
$real_global_pool_settings = $global_pool_settings
$real_pools = $pools
```
this solution does not need much effort. the cleaner way would be to refactor the used variable names.


i already fixed that in my fork quite a while ago for `init.pp` https://github.com/c33s/puppet-php/commit/0363ab3db8fb490fa6baa787b3bc602e33d551fa and the othe fix was merged here https://github.com/voxpupuli/puppet-php/pull/435/files


@sigv & @Dan33l i am not sure (can't remember) if it simply works by adding the `lookup_options` does they also change the behavior for the `lookup()` function with hardcoded options? the reason for my fork was that with the hardcoded lookup functions it was **not** possible for me to change the merge behavior for these functions, it was only for the automatic parameter lookup. maybe this has changed in higher puppet versions?

for me using the lookup function inline breaks puppet. the remove of this function would be the right way for me.

sigv:
@c33s Fair. Dropped `lookup` function entirely for the `fpm` here.

data/default.yaml
```diff
@@ -1,5 +1,5 @@
 ---
```
c33s: shouldn't we add the correct setting for the merge behavior here? your target was to get `deep` as merge behavior?

@bastelfreak what whould be a good default? i personally prefer `first` over `deep` for pools. don't want my pools get collected from multiple files but for other persons `deep` is maybe the better default. so what to choose?

@sigv what would be your prefered default?

```
lookup_options:
  php::fpm::pools:
    merge: deep
```

Dan33l:
> @sigv & @Dan33l i am not sure (can't remember) if it simply works by adding the `lookup_options` does they also change the behavior for the `lookup()` function with hardcoded options? the reason for my fork was that with the hardcoded lookup functions it was **not** possible for me to change the merge behavior for these functions, it was only for the automatic parameter lookup. maybe this has changed in higher puppet versions?

My understanding is also that `lookup_options` key in hiera works only with automatic parameter lookup. 

I apologize, i though we were talking about automatic parameter lookup. I am not a user of this module. So, now i read the code, i am seeing that it is hard coded lookup.

I agree that a clean way is to find a way to not hard code the lookup, if possible.

data/default.yaml
Dan33l: If the choice of merge strategy is so important, we probably need a note about this and how to change it in the `README.md` file.

sigv:
> I agree that a clean way is to find a way to not hard code the lookup, if possible.

That is indeed the current state of the PR - to switch the FPM pools variables over to automatic lookup.

data/default.yaml
sigv: The option to choose other strategies is a standard Puppet feature. For example, `php::extensions` can already be configured to merge in different ways - it's also a Hash being taken from Hiera. The README does not have any documentation about changing merge behavior for that variable. Additionally, [automatic lookup documentation](https://puppet.com/docs/puppet/6.4/hiera_automatic.html#class-parameters) does cover the fact hashes are `first` by default:

> Note: Automatic lookup of class parameters uses the "first" merge method by default. You cannot change the default. If you want to get deep merges on keys, use the lookup_options feature.

Should I really add a note about this in this module's README in this PR?

sigv: Just pushed a minor note in `README` as commit 5d59134 in this PR. Please check that out and let me know if further changes to this PR are of interest. Otherwise, I would prefer a _LGTM_ so that the change itself gets merged. 🙂 🙈

c33s: readme looks good to me. 

i miss the following addition of ` data/default.yaml`:
```
lookup_options:
  php::fpm::pools:
    merge: first
```
i also would reference this file in the readme, so user see the context of the config. after that i would approve the MR

sigv: 👌🏻 Pushed `hash` as the override in `data/default.yaml` simply because `first` is already the default and overriding to it doesn't make that much sense. Added a link to the README pointing at `data/default.yaml` as well.

c33s: can you please use `first`? my request to put it there was just because of having a reference what to change if someone want to change the behavior but i would prefere to stick with the default setting of `first`

sigv: Done. It is set to `first` now in the local Hiera source.

sigv: Ping.

sigv:
Is this still waiting for further approval after the LGTM? I don't have push/merge rights.

c33s:
> Is this still waiting for further approval after the LGTM? I don't have push/merge rights.

we are waiting for @bastelfreak s approval. he's quite busy right now. i have the permission to merge, but it is my first merge here, so i wait for his decision. after he is ok with my reviews i will merge such requests faster in future.

sigv:
Got it - no worries. Just wanted to hear what's going on.

bastelfreak:
Hi people, sorry for the delay here.

phyber:
What's the actual fix here? It's not very clear at all, and the suggestion in the documentation still results in this error.

c33s:
@phyber what exactly are you trying to do? can you show me some code snippets?

you only have to add the following code to your hiera data for example `common.yaml`
```yaml
lookup_options:
  php::fpm::pools:
    merge: hash
```
my `hieradata/common.yaml` looks like that:

```yaml
lookup_options:
  profile::nginx::servers: { merge: {strategy: deep, merge_hash_arrays: true, knockout_prefix: "--", sort_merged_arrays: true}}
  profile::nginx::upstreams: { merge: {strategy: deep, merge_hash_arrays: true, knockout_prefix: "--", sort_merged_arrays: true}}
  profile::nginx::mailhost: { merge: {strategy: deep, merge_hash_arrays: true, knockout_prefix: "--", sort_merged_arrays: true}}
  sysctl::values: { merge: first }
```

phyber:
The version of Puppet here is 5.5.10 (Debian stable), the PHP module is 7.0.0.

I have `php::fpm::pools` defined for a single node under its own branch in hiera, `nodes/%{::hostname}.yaml`. I only have a `php::fpm::pools` defined for this single node and nowhere else in the hiera tree. The hash is pretty standard, with no lookups or anything.
```yaml
php::fpm::pools:
    www:
        user: 'www-data'
        group: 'www-data'
        listen: '/run/php/php7.3-fpm.sock'
        etc: ...
```

The lookup options are specified in my top level `common.yaml` in long form.
```yaml
lookup_options:
    php::fpm::pools:
        merge:
            strategy: 'hash'
```

We can see this working in the `puppet lookup --explain`:

```
# puppet lookup --explain php::fpm::pools
<some truncated output>
Using merge options from "lookup_options" hash
Searching for "php::fpm::pools"
  Merge strategy hash
<truncated output>
    Merged result: {
      "www" => {
        "user" => "www-data",
        "group" => "www-data",
        "listen" => "/run/php/php7.3-fpm.sock",
        "etc" => ...,
      }
    }
```

Running apply or noop results in the following:
```
# make noop
/usr/bin/puppet apply --noop /etc/puppet/code/environments/production/manifests/site.pp
Error: Evaluation Error: Error while evaluating a Function Call, Found value has wrong type, expects a Hash value, got Tuple (file: /etc/puppet/code/environments/production/vendor/modules/php/manifests/fpm.pp, line: 103, column: 17) on node <hostname>
```

Everything looks like it _should_ be working.

phyber:
After playing around with this, it feels like what's happening here is that it doesn't matter that you passed `strategy: hash` via `lookup_options` in hiera.

This appears to be because the merge strategy in the `lookup` function call (`strategy: unique`) is invalid for the type of values being passed (hashes) at the time the function call is made.
The overriding of the lookup type via hiera `lookup_options` probably happens within the `lookup` function call later on, but by then, the function has already errored out. So no matter what, this appears to be broken until the fix is released.

Fortunately, it looks like 7.0.1 is happening soon (#563), which should fix the issue thanks to this PR.

PR #552, opened 2019-11-04
Improve package prefix selection on FreeBSD

oxc:
On FreeBSD, the package prefix is currently hardcoded to php56. This change modifies the default behaviour to use the prefix matching the selected php version (from globals) if defined.

This change should be fully compatible, since the default is to fallback to php56, and if a custom version is specified, the package prefix has to be manually configured to match this version anyway.

Futhermore, the pear package is now using a php prefix as well (since 2018-03-08), which is also fixed by this pull request.

### DIFF
--- a/manifests/globals.pp
+++ b/manifests/globals.pp
@@ -144,6 +144,14 @@ class php::globals (
       }
     }
     'FreeBSD': {
+      case $globals_php_version {
+        /^(\d)\.(\d)$/: {
+          $package_prefix = "php${1}${2}-"
+        }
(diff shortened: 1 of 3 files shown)


ghoneycutt:
Tests would help this get merged or someone else verifying this works with FreeBSD 9-11.

bastelfreak:
@oxc  please rebase against our latest master branch.

oxc:
> @oxc please rebase against our latest master branch.

Done.

PR #554, opened 2019-12-10
Fix running apt update on Debian family

Hexta:
Fix os::family name for Debian.

Fix running apt update on Debian family.
Replace 'debian' by 'Debian' in family check in accordance with Facter family name constants: https://github.com/puppetlabs/facter/blob/2aa2d1cd6487f73fb8e311cf027dcf30c2166e28/lib/inc/facter/facts/os_family.hpp

### DIFF
--- a/manifests/packages.pp
+++ b/manifests/packages.pp
@@ -22,7 +22,7 @@ class php::packages (
   assert_private()
 
   $real_names = union($names, $names_to_prefix)
-  if $facts['os']['family'] == 'debian' {
+  if $facts['os']['family'] == 'Debian' {
     if $manage_repos {
       include ::apt
       Class['::apt::update'] -> Package[$real_names]


c33s:
@Hexta thank you very much for your pull request. just for the records do you had a problem with the lowercase code? @bastelfreak wrote with me that puppet is case insentive in most cases. does the apt part not executed for you?

c33s:
@Hexta can you add a test to verify that the section is now correctly called if the os is `Debian`?

Hexta:
> @Hexta thank you very much for your pull request. just for the records do you had a problem with the lowercase code? @bastelfreak wrote with me that puppet is case insentive in most cases. does the apt part not executed for you?

Sorry, the issue was actually fixed in https://github.com/voxpupuli/puppet-php/pull/555

PR #555, opened 2019-12-10
Allow sury repo for PHP >= 7.1.

Hexta:
Allow sury repo for PHP >= 7.1.

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -67,7 +67,13 @@ class php::repo::debian(
     }
   }
 
-  if ($sury and $php::globals::php_version in ['5.6','7.1','7.2'] ) {
+  $php_version = $php::globals::php_version
+
+  if ($sury and $php_version != undef and
+      (
+        $php_version == '5.6'
+        or versioncmp($php_version, '7.1') >= 0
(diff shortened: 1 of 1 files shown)


Hexta:
Could anybody please  take a look at the PR?

manifests/repo/debian.pp
```diff
@@ -67,7 +67,13 @@
     }
   }
 
-  if ($sury and $php::globals::php_version in ['5.6','7.1','7.2'] ) {
```
c33s: would be better to simply add the supported php version of sury in the array, only this way non valid php versions could be catched.
with your change a php version like `8.0` would be allowed

Hexta: Do you really think that hardcoded list of supported php versions is a good idea?

c33s: not really, i prefer to have them configureable via hiera with sane default values directly set in this module with hiera. i would prefer that for many hardcoded values (like the repo key id) but without a bigger refactoring this is not possible.

to simply remove the version check feels wrong to me.

bastelfreak:
@Hexta please rebase against our latest master branch.

PR #565, opened 2020-02-13
Add validation of fpm and fpm-pools configuration files

ffapitalle:

This PR adds validation check for fpm and fpm pools configuration file before applying changes.

Having introduced a misconfiguration on php-fpm that avoid the service to start normally (`pm.start_servers` > `pm.max_spare_servers`), catalog application was successfull but service failed to restart.

```
[13-Feb-2020 20:48:24] ALERT: [pool www] pm.start_servers(80) must not be less than pm.min_spare_servers(24) and not greater than pm.max_spare_servers(64)
[13-Feb-2020 20:48:24] ERROR: failed to post process the configuration
```


manifests/fpm/config.pp
```diff
@@ -103,12 +103,18 @@
     default => $log_group,
   }
 
+  $validate_cmd = $php::fpm::disable_configtest ? {
```
ghoneycutt: Why wouldn't you want to validate this? Not sure the boolean is needed.

ghoneycutt:
Thanks for the contribution! Could you please add an acceptance test.

manifests/fpm.pp
```diff
@@ -64,6 +64,7 @@
   Hash $pools                   = $php::real_fpm_pools,
   $log_owner                    = $php::log_owner,
   $log_group                    = $php::log_group,
+  Boolean $disable_configtest   = false,
```
ghoneycutt: This would need a spec test to show that `validate_cmd` is as expected.

manifests/fpm/config.pp
ffapitalle: In my use case, I want validation always. But, I took the idea from [auditbeat's module](https://forge.puppet.com/norisnetwork/auditbeat), which gives the opportunity to maintain the behaviour as today, being the default value that enables validation.

manifests/fpm.pp
ffapitalle: Yes, indeed. I'll add a spec test for this parameter.

PR #570, opened 2020-03-29
Cleanup fpm config class

### LINKED ISSUE #568: Unused variable 'log_group_final'

- Puppet: irrelevant
- Ruby: irrelevant
- Distribution: irrelevant
- Module version: v7.0.0

irrelevant

## What are you seeing

The variable 'log_group_final', defined in [fpm/config.pp](https://github.com/voxpupuli/puppet-php/blob/v7.0.0/manifests/fpm/config.pp#L100-L104), is not used anywhere.

```
~/puppet-php $ grep log_group_final -r
docs/puppet_classes/php_3A_3Afpm_3A_3Aconfig.html:  $log_group_final = $log_group ? {
manifests/fpm/config.pp:  $log_group_final = $log_group ? {
```

The variable assignment should be removed.

paescuj:
This pull request cleans up the fpm config class:
- Remove unused variable 'log_group_final'
- Fixes owner, group and mode for the directories '/var/log/php-fpm' and '/var/run/php-fpm'

/var/log/php-fpm:
We don't know which processes want to access this directory - therefore root with mode 0755 is used here. This is also the default within most distributions.

/var/run/php-fpm:
There is no need to set the PHP-FPM user/group as a owner of this directory. This is also follows the default within most distributions.

Fixes #568

### DIFF
--- a/manifests/fpm/config.pp
+++ b/manifests/fpm/config.pp
@@ -97,29 +97,31 @@ class php::fpm::config(
 
   assert_private()
 
-  # Hack-ish to default to user for group too
-  $log_group_final = $log_group ? {
-    undef   => $log_owner,
-    default => $log_group,
-  }
-
   file { $config_file:
     ensure  => file,
     content => template('php/fpm/php-fpm.conf.erb'),
-    owner   => root,
+    owner   => 'root',
     group   => $root_group,
     mode    => '0644',
   }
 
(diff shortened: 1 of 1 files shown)


bastelfreak:
@paescuj please rebase against our latest master branch.

paescuj:
@bastelfreak Rebase is done!

PR #571, opened 2020-04-09
add ability to define composer update channel

CyberLine:

Yesterday, composer released 2.0 snapshot.
So with current default behaviour, if i only set

`class { '::php::composer':
    max_age => 2,
}`

every composer installation is updated to version 2.0 snapshot.

So one solution is to set the 'source' to 'https://getcomposer.org/composer-stable.phar' initialy if i dont have it installed or use the param from this PR for updates, which then defaults update channel to 'stable'

### DIFF
--- a/manifests/composer.pp
+++ b/manifests/composer.pp
@@ -8,6 +8,9 @@
 # [*path*]
 #   Holds path to the Composer executable
 #
+# [*channel*]
+#   Holds the Update channel (stable|preview|snapshot|1|2)
+#
 # [*proxy_type*]
 #    proxy server type (none|http|https|ftp)
 #
@@ -28,6 +31,7 @@ class php::composer (
   Stdlib::Absolutepath $path           = $php::params::composer_path,
   $proxy_type                          = undef,
   $proxy_server                        = undef,
+  Php::ComposerChannel $channel        = 'stable',
   Boolean $auto_update                 = true,
   Integer $max_age                     = $php::params::composer_max_age,
   Variant[Integer, String] $root_group = $php::params::root_group,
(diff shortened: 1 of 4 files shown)


CyberLine:
Since Composer 1.10.2 today, --1 and --2 are also possible channel names.

c33s:
as far as i can see there are some changed lines which contain only whitespace changes. can you please ensure that there are only clean diffs of lines you really changed. this helps with the review.

manifests/composer.pp
```diff
@@ -28,6 +31,7 @@
   Stdlib::Absolutepath $path           = $php::params::composer_path,
   $proxy_type                          = undef,
   $proxy_server                        = undef,
+  $channel                             = 'stable',
```
bastelfreak, line 34: Please add a datatype for this. Would an enum make sense here?

CyberLine: Hi Tim, i've added it as an enum now.

c33s: looks good to me. only the test fixes are left.

CyberLine:
Tests are working now except https://travis-ci.org/github/voxpupuli/puppet-php/jobs/673719739 and https://travis-ci.org/github/voxpupuli/puppet-php/jobs/673719740 which are not related to this PR.

bastelfreak:
@CyberLine please rebase after https://github.com/voxpupuli/puppet-php/pull/513/ got merged

manifests/composer.pp
```diff
@@ -28,6 +31,7 @@
   Stdlib::Absolutepath $path           = $php::params::composer_path,
   $proxy_type                          = undef,
   $proxy_server                        = undef,
+  Php::ComposerChannel $channel      = 'stable',
```
bastelfreak, line 34:
```suggestion
  Php::ComposerChannel $channel       = 'stable',
```

Issue #558, opened 2019-12-17
fpm_pools option not working after release 7.0.0


bovy89:
After [this](https://github.com/voxpupuli/puppet-php/commit/75c5e4e731e6ea9fce120c4c14f8ee8b8d4fca2b) change, fpm_pools does not work anymore.

As consequence, [this](https://github.com/voxpupuli/puppet-php#php-with-one-fpm-pool-per-user) use case is broken too

Version 7.0.0:

```
class { 'php':
   ensure       => 'present',
   manage_repos => false,
   fpm          => true,
   dev          => false,
   composer     => false,
   pear         => true,
   phpunit      => false,
   fpm_pools    => {},
}

php::fpm::pool { 'www2':
    listen => '127.0.1.1:9000',
}
```

Created pools: www2 ---> correct

Afred 7.0.0:

```
class { 'php':
   ensure       => 'present',
   manage_repos => false,
   fpm          => true,
   dev          => false,
   composer     => false,
   pear         => true,
   phpunit      => false,
   fpm_pools    => {},
}

php::fpm::pool { 'www2':
    listen => '127.0.1.1:9000',
}

```

Created pools: www and www2 ----> wrong

jcharaoui:
I can confirm this unexpected behavior. Parameter `fpm_pools` in the base class appears not to have any effect anymore.

grant-veepshosting:
Perhaps this could be addressed by adding a knockout_prefix for the default www pool?

@jcharaoui if you refer to the linked commit above, the parameter 'php::fpm_pools' has moved to 'php::fpm::pools' instead, you'd need to change your Hiera data to reflect this.

jouveits:
Furthermore, the default 'www' pool cannot be customized either using the php class. 

For instance, this has no effect:

```php
class {'::php':
  # ...
  fpm_pools => {
    'www' => {
      user => 'php',
      pm   => 'static'
    }
  }
}
```

cedriclaudrel:
Can confirm, it's currently impossible to unset the default `www` pool without using Hiera or relying on overriding private classes parameters.

bovy89:
This issue has been resolved in master (see #572), but a new release is required

PR #557, opened 2019-12-17
add pool_purge option to init.pp

bovy89:
Similar to the following stale PR: https://github.com/voxpupuli/puppet-php/pull/515/files

PR #572, opened 2020-04-27
remove www pool from defaults

bovy89:
This PR fixs #558 and let #557 to work as intended


StanvanHoorn:
Hi, 

Any idea when this is added to a new release?

PR #576, opened 2020-05-12
Download stable version of composer as opposed to snapshot

### LINKED ISSUE #575: Snapshot composer is installed instead of stable composer

- Puppet: 6.14.0
- Ruby: 2.7
- Distribution: Centos7
- Module version: 7.1.0

Should the composer source parameter (https://github.com/voxpupuli/puppet-php/blob/master/manifests/params.pp) be updated to download the stable version of composer? Since the default channels are stable as well? Currently, it's downloading the snapshot version which breaks some installs.

this:

    $composer_source     = 'https://getcomposer.org/composer.phar'

to this: 

    $composer_source     = 'https://getcomposer.org/composer-stable.phar'

as listed on https://getcomposer.org/download/

plexium:

Download stable version of composer as opposed to snapshot

Fixes #575

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,7 +5,7 @@ class php::params inherits php::globals {
   $ensure              = 'present'
   $fpm_service_enable  = true
   $fpm_service_ensure  = 'running'
-  $composer_source     = 'https://getcomposer.org/composer.phar'
+  $composer_source     = 'https://getcomposer.org/composer-stable.phar'
   $composer_path       = '/usr/local/bin/composer'
   $composer_max_age    = 30
   $pear_ensure         = 'present'


bastelfreak:
Is this a backwards incompatible change? The fold url is still valid, just a different composer version?

dhoppe:
> Is this a backwards incompatible change? The fold url is still valid, just a different composer version?

In my opinion this looks more like a bug fix:

> By default the installer and composer self-update will download stable versions only. If you would like to help test pre-release versions you can use the --preview flag on either installer or self-update. For snapshot builds which are done from the latest Composer commit, you can use the --snapshot flag.

plexium:
> Is this a backwards incompatible change? The fold url is still valid, just a different composer version?

Correct, it's still a valid url. I noticed this being a problem when provisioning a new node, the composer version that would install was a dev version, and caused a bunch of composer installs downstream in my manifests to fail. I would have to manually do a `composer selfupdate` after provisioning to get the stable version. 

Since is saw the [update branch defaults to the 'stable'](https://github.com/voxpupuli/puppet-php/blob/master/manifests/composer/auto_update.pp#L35), so should the initial install.

I don't think it's a backwards incompatible change.

yakatz (approved):

PR #590, opened 2020-07-31
modulesync 3.0.0 & puppet-lint autofix

bastelfreak:


bastelfreak:
fixes https://github.com/voxpupuli/puppet-php/issues/591

bastelfreak:
I didn't mention freebsd here because that's handled in the issue https://github.com/voxpupuli/puppet-php/issues/591. this seperation makes it cleaner in the changelog

Issue #596, opened 2020-08-26
Restart service 'php-fpm' if socket owner / group changes


kBite:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.18.0
- Ruby: 2.5.5p157
- Distribution: CentOS 8
- Module version: v7.1.0

## How to reproduce (e.g Puppet code you use)

In Hiera
```
php::fpm::service_enable: 'running'   
php::fpm::service_enable: true        
php::fpm::config::user: 'apache'      
php::fpm::config::group: 'apache'     
php::fpm::pools:                        
  test-pool:              
    user: 'apache'                      
    group: 'apache'                     
    listen: '/run/php-fpm/$pool.sock'   
    listen_owner: 'apache'              
    listen_group: 'apache' 
```             

## What are you seeing

Puppet run triggers only a reload which does not change the sockets owner and group
```
Debug: Executing: 'service php-fpm reload'
Notice: /Service[php-fpm]: Triggered 'refresh' from 1 event
```

```
# ls -l /run/php-fpm/test-pool.sock                                     
srw-rw---- 1 root root 0 26. Aug 18:18 /run/php-fpm/test-pool.sock      
                                                                        
# systemctl reload php-fpm                                              
                                                                        
# ls -l /run/php-fpm/test-pool.sock                                     
srw-rw---- 1 root root 0 26. Aug 18:18 /run/php-fpm/test-pool.sock      
                                                                        
# systemctl restart php-fpm                                             
                                                                        
# ls -l /run/php-fpm/test-pool.sock                                     
srw-rw---- 1 apache apache 0 26. Aug 18:31 /run/php-fpm/test-pool.sock  
```                                                                    

## What behaviour did you expect instead

Seeing changed owner / group right after Puppet run without manual restarting
```
# ls -l /run/php-fpm/test-pool.sock                                     
srw-rw---- 1 apache apache 0 26. Aug 18:31 /run/php-fpm/test-pool.sock
```

## Any additional information you'd like to impart

This behaviour is hardcoded in `fpm/service.pp`

```
$reload = "service ${service_name} reload"

if ($facts['os']['name'] == 'Ubuntu'
    and versioncmp($facts['os']['release']['full'], '12') >= 0
    and versioncmp($facts['os']['release']['full'], '14') < 0) {
  # Precise upstart doesn't support reload signals, so use
  # regular service restart instead
  $restart = undef
} else {
  $restart = $reload
}

service { $service_name:
[...]

PR #597, opened 2020-08-27
(#596) drop 'reload' in favour of 'restart'

kBite:
#### Pull Request (PR) description

Some configuration options (e.g. 'listen_owner', 'listen_group', 'listen_mode', and maybe some more) do not apply by reloading 'php-fpm.service' but require a proper restart.

#### This Pull Request (PR) fixes the following issues

Fixes #596

PR #598, opened 2020-08-29
php-fpm: Make service reload/restart configurable

bastelfreak:
By default, this module uses a reload for the php-fpm service. This is
nice because it ensures that there is no downtime. But this is also bad
because certain configuration options, like unix socket owner/group will
only be changed during a restart.

Fixes: https://github.com/voxpupuli/puppet-php/issues/596
Replaces: https://github.com/voxpupuli/puppet-php/pull/597

### DIFF
--- a/manifests/fpm.pp
+++ b/manifests/fpm.pp
@@ -53,22 +53,29 @@
 #   Whether to purge pool config files not created
 #   by this module
 #
+# [*reload_fpm_on_config_changes*]
+#   by default, we reload the service on changes.
+#   But certain options, like socket owner, will only be applied during a restart.
+#   If set to false, a restart will be executed instead of a reload.
+#   This default will be changed in a future release.
+#
 class php::fpm (
-  String $ensure                = $php::ensure,
-  $user                         = $php::fpm_user,
-  $group                        = $php::fpm_group,
-  $service_ensure               = $php::fpm_service_ensure,
-  $service_enable               = $php::fpm_service_enable,
-  $service_name                 = $php::fpm_service_name,
-  $service_provider             = $php::fpm_service_provider,
-  String $package               = $php::real_fpm_package,
-  Stdlib::Absolutepath $inifile = $php::fpm_inifile,
-  Hash $settings                = $php::real_settings,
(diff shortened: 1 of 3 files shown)


c33s (approved):
lgtm

manifests/fpm.pp
```diff
@@ -53,22 +53,29 @@
 #   Whether to purge pool config files not created
 #   by this module
 #
+# [*reload_fpm_on_config_changes*]
+#   by default, we reload the service on changes.
+#   But certain options, like socket owner, will only be applied during a restart.
+#   If set to false, a restart will be executed instead of a reload.
+#   This default should be changed in a future release.
```
igalic, line 60: should be ➡️ will be

manifests/init.pp
```diff
@@ -127,6 +127,12 @@
 #   Whether to purge pool config files not created
 #   by this module
 #
+# [*reload_fpm_on_config_changes*]
+#   by default, we reload the service on changes.
+#   But certain options, like socket owner, will only be applied during a restart.
+#   If set to false, a restart will be executed instead of a reload.
+#   This default should be changed in a future release.
```
igalic, line 134: should be ➡️ will be

manifests/fpm/service.pp
```diff
@@ -14,21 +14,34 @@
 # [*provider*]
 #   Defines if the service provider to use
 #
+# [*reload_fpm_on_config_changes*]
+#   by default, we reload the service on changes.
+#   But certain options, like socket owner, will only be applied during a restart.
+#   If set to false, a restart will be executed instead of a reload.
+#   This default should be changed in a future release.
```
igalic, line 21: should be ➡️ will be

igalic (changes requested):
👀

manifests/fpm/service.pp
bastelfreak: fixed

manifests/init.pp
bastelfreak: fixed

manifests/fpm.pp
bastelfreak: fixed

PR #601, opened 2020-09-26
Adding the option to set process.dumpable

motivator:
https://code.bigcommerce.net/techops/puppet/-/merge_requests/10894

This puts back our custom option. The MR ^^ has the info.


igalic:
is there a particular reason you opened (and closed) three of these?

PR #602, opened 2020-09-26
Adding the option to set process.dumpable

motivator:
https://code.bigcommerce.net/techops/puppet/-/merge_requests/10894

This puts back our custom option. The MR ^^ has the info.


igalic:
> The connection has timed out

one of the reasons i prefer having the info in here

PR #605, opened 2020-09-27
WIP: Feature/simon hoenscheid 603 restructure repository management

### LINKED ISSUE #603: [Rewrite] Restructure Repository Management
This Ticket describes what the new Repository Management looks like

* ideally backwards compatible
* all repo data from hiera (data in modules)
* clear and easy to understand conditions
* easy to add, update and remove repos

TODO: 

* create a list/matrix of supported PHP versions and operating systems (seperate issue)
* create a list of non-default repos
* create logic
* create data
* write code
* write tests
* merge to module_rewrite branch

SimonHoenscheid:
Fixes #603


data/CentOS/6.yaml
```diff
@@ -0,0 +1 @@
+---
```
bastelfreak, line 1: can you please add the missing `\n` at the end of this line? (same applies for the other files)

bastelfreak: in the past we also had issues with empty yaml files, but I cannot remember what broke and which conditions :(

manifests/repo/debian.pp
```diff
@@ -2,66 +2,18 @@
...
-  $release      = 'wheezy-php56',
-  $repos        = 'all',
-  $include_src  = false,
-  $key          = {
-    'id'     => '6572BBEF1B5FF28B28B706837E3F070089DF5277',
-    'source' => 'http://www.dotdeb.org/dotdeb.gpg',
-  },
-  $dotdeb       = true,
-  $sury         = true,
+  Hash $external_repo_details = $php::external_repo_details,
```
bastelfreak, line 9: can we make the type validation more strict with this?
```suggestion
  Hash[String[1], Hash] $external_repo_details = $php::external_repo_details,
```

manifests/repo/fallback.pp
```diff
@@ -0,0 +1,7 @@
+# the fallback class, throws an error, if called
```
bastelfreak, line 1:
```suggestion
# @summary the fallback class, throws an error, if called
```

also please add a `# @api private` to the header

manifests/repo.pp
```diff
@@ -1,32 +1,15 @@
...
-      fail($msg_no_repo)
-    }
+class php::repo (
+  Boolean $external_repo                                = $php::external_repo,
+  Optional[String[1]] $php_version                      = $php::php_version,
+  Optional[Array] $external_repo_supported_php_versions = $php::external_repo_supported_php_versions,
+) {
+  if ($external_repo and $php_version in $external_repo_supported_php_versions) {
+  case $facts['os']['name'] {
+  'RedHat', 'CentOS': { contain '::php::repo::redhat'  }
```
bastelfreak, line 10: as far as I know, calling classes or variables by top scope is a bad pattern and should be avoided:
```suggestion
  'RedHat', 'CentOS': { contain 'php::repo::redhat'  }
```

manifests/repo.pp
```diff
@@ -1,32 +1,15 @@
...
+  'Debian', 'Ubuntu': { contain '::php::repo::debian'  }
+  default:            { contain '::php::repo::fallback' }
```
bastelfreak, line 12:
```suggestion
  default:            { contain 'php::repo::fallback' }
```

manifests/repo.pp
```diff
@@ -1,32 +1,15 @@
...
-    }
+class php::repo (
+  Boolean $external_repo                                = $php::external_repo,
+  Optional[String[1]] $php_version                      = $php::php_version,
+  Optional[Array] $external_repo_supported_php_versions = $php::external_repo_supported_php_versions,
+) {
+  if ($external_repo and $php_version in $external_repo_supported_php_versions) {
+  case $facts['os']['name'] {
+  'RedHat', 'CentOS': { contain '::php::repo::redhat'  }
+  'Debian', 'Ubuntu': { contain '::php::repo::debian'  }
```
bastelfreak, line 11:
```suggestion
  'Debian', 'Ubuntu': { contain 'php::repo::debian'  }
```

manifests/init.pp
```diff
@@ -133,6 +133,33 @@
 #   If set to false, a restart will be executed instead of a reload.
 #   This default will be changed in a future release.
 #
+#########
+######### REWRITE
+#
+# [*external_repo*]
```
bastelfreak, line 139: I think we should use puppet-strings style documentation for new parameters and not javadoc (or however this is called)

SimonHoenscheid: did I miss a style change?

hiera.yaml
```diff
@@ -1,6 +1,13 @@
 ---
 version: 5
+defaults:
+  datadir: 'data'
+  data_hash: 'yaml_data'
+
 hierarchy:
-  - name: default.yaml
+  - name: 'My Hierachie'
```
bastelfreak, line 8:
```suggestion
  - name: 'My Hierarchies'
```

manifests/repo.pp
```diff
@@ -1,32 +1,53 @@
...
-      contain 'php::repo::redhat'
-    }
-    default: {
-      fail($msg_no_repo)
+) {
+  assert_private()
+
+  if (($manage_internal_repo or $manage_external_repo) and (($php_version in $external_repo_supported_php_versions) or ($php_version in $os_supported_php_versions))) {
+    case $facts['os']['name'] {
+      'Archlinux':                  { contain 'php::repo::archlinux' }
```
bastelfreak, line 45: for arch and freebsd, should we just do nothing? since those don't have external repositories.
```suggestion
      'Archlinux':                  { }
```

SimonHoenscheid: The classes are for generally managing repos, is there no need to connect a specific one, for a specific php version? Kind of the same problem exists for RHEL style distros, there is the need to activate a specific app stream

manifests/init.pp
```diff
@@ -169,6 +206,15 @@
   String $log_group                               = $php::params::fpm_group,
   Boolean $pool_purge                             = $php::params::pool_purge,
   Boolean $reload_fpm_on_config_changes           = true,
+  # added for refactoring
+  Boolean $manage_external_repo                         = false,
+  Array $external_repo_supported_php_versions           = [],
+  Hash $external_repo_details                           = {},
```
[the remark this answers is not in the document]
SimonHoenscheid: does this work for every repo type?

manifests/repo.pp
```diff
@@ -1,32 +1,53 @@
...
+# @option os_supported_php_versions [Array] :[]
+#   default: empty array, default for OS from Hiera Data in module
+# 
+# @param [String] php_version
+#   Set the PHP version to install
+# @option php_version [String] :'7.4'
+#   defaults: latest stable PHP Version available on php.net, default for OS from Hiera Data in module
+#
+class php::repo (
+  Boolean $manage_internal_repo                         = $php::manage_internal_repo,
```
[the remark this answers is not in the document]
SimonHoenscheid: In my opinion yes, because there are operating systems where we need to manage internal and external repos (RHEL/CENTOS/Fedora)  app channel/Remi and if I want to have different data without interpolation this seems to be the way to go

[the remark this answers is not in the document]
c33s: don't know about centos & co, you manage the internal and the external repo for php?

is it relevant if it is external or internal? isn't it just multiple `repositories`?

manifests/init.pp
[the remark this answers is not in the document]
c33s: "yes" we would have to write a type for each repo stucture we want to support.

manifests/repo.pp
SimonHoenscheid: Setz ich so um

manifests/init.pp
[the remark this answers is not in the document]
SimonHoenscheid: I will add this type to the module

manifests/repo.pp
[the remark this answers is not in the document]
SimonHoenscheid: Its been a while since I worked with Fedora based systems, it only affects RHEL8/CENTOS8 and Fedora IDK which version RHEL8 was branched from +. For me this seems important, because it will prevent users declaring internal and external php versions/sources parallel.

SimonHoenscheid:
I would now remove all old tests so we can rewrite the tests for rewritten classes. So the test status is clear

PR #610, opened 2020-11-25
modulesync 4.0.0

### LINKED ISSUE #616: remove puppet 5 support, introduce puppet 7 support
this module currently supports a old puppet matrix this should be brought into line with other modules

### LINKED ISSUE #617: update supported os matrix
this module currently supports an old matrix of supported OS-es this should be brought in line with other modules

bastelfreak:
modulesync 4.0.0

Fixes #616
Fixes #617


ekohl (changes requested):
It fails acceptance tests because previously it split everything into 2 sets: `php_spec.rb` and `php56_spec.rb`. That needs to be codified into GHA now.

attachmentgenie:
@bastelfreak i am happy to look into getting this PR across the line, do you mind me cleaning up the supported matrix at the same time?

attachmentgenie:
the problem is the current setup seems to run all acceptance tests in the same container, which causes problem with artifacts from the php56 test still being there when running the next test. Is this how the acceptance beaker setup is supposed to work or are the containers meant to be cleaned up after every individual test? i am not entirely sure how this is supposed to work in beaker /cc @ekohl

kenyon:
> 
> 
> the problem is the current setup seems to run all acceptance tests in the same container, which causes problem with artifacts from the php56 test still being there when running the next test. Is this how the acceptance beaker setup is supposed to work or are the containers meant to be cleaned up after every individual test? i am not entirely sure how this is supposed to work in beaker /cc @ekohl

Yes, that's how it works, no cleanup is done for you. So what I've done before is a sort of manual cleanup, here is an example: https://github.com/voxpupuli/puppet-nodejs/blob/master/spec/acceptance/class_spec.rb#L83-L99

metadata.json
```diff
@@ -55,15 +54,15 @@
     {
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
-        "6",
-        "7"
+        "7",
+        "8"
```
ekohl, line 58: Not a problem, but we shouldn't forget to make issues for this so it's clear in the changelog.

metadata.json
```diff
@@ -33,14 +33,13 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 7.0.0"
+      "version_requirement": ">= 6.1.0 < 8.0.0"
```
ekohl, line 36: Not a problem, but we shouldn't forget to make issues for this so it's

ekohl (commented):
For now I've pushed a change that runs the tests sequentially and I think that it should tear down the old env first and then run the new env. I know that @trevor-vaughan has done some work with test tiers. This is something we should probably also include in our test envs.

metadata.json
attachmentgenie: ok will do

metadata.json
attachmentgenie: will do

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+      - uses: actions/checkout@v2
+      - name: Setup ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '2.7'
+          bundler-cache: true
+      - name: Run tests
+        run: |
+          bundle exec rspec spec/acceptance/php_spec.rb
+          bundle exec rspec spec/acceptance/php56_spec.rb
```
ekohl, line 74: This now fails on some combinations so we need to find a way to skip this. We can either skip it here or add conditionals in php56_spec.rb. I'm leaning to the latter.

attachmentgenie:
@ekohl thanks for showing how to split the runs into distinct containers. 

As a side node or related questions how much effort should i put in the make the php 5.6 code work. since php 5.6 has been EOL since 31 Dec 2018. To me this smells similar to the centos6 and puppet 5 conversations we have been having

/cc @bastelfreak @kenyon

ekohl:
I think the main goal is to test the custom repo code, not specifically version 5.6.

The Debian repo uses dotdeb.org but that's been offline for a while. See https://github.com/voxpupuli/puppet-php/blob/master/manifests/repo/debian.pp. The Ubuntu PPA looks to be maintained, which is why it's passing. He also appears to maintain Debian packages so that may be the easiest switch. Then just skip the test on EL8 or choose a newer version that is provided.

attachmentgenie:
> I think the main goal is to test the custom repo code, not specifically version 5.6.
> 
> The Debian repo uses dotdeb.org but that's been offline for a while. See https://github.com/voxpupuli/puppet-php/blob/master/manifests/repo/debian.pp. The Ubuntu PPA looks to be maintained, which is why it's passing. He also appears to maintain Debian packages so that may be the easiest switch. Then just skip the test on EL8 or choose a newer version that is provided.

i started to go down this rabbit hole to get php 8 working so i might just switch to that version that to prove working code

ekohl:
Sounds good. Also feel free to rename the spec to `custom_repo_spec.rb` to make it clear what the intention is.

manifests/repo/debian.pp
```diff
@@ -2,66 +2,29 @@
...
-        'src' => $include_src,
-        'deb' => true,
-      },
-      key      => {
-        id     => '15058500A0235D97F5D10063B188E2B695BD4743',
-        source => 'https://packages.sury.org/php/apt.gpg',
-      },
-    }
+    key      => {
+      id     => '15058500A0235D97F5D10063B188E2B695BD4743',
```
ekohl, line 26: Hardcoding keys is IMHO a bad practice. If the key changes, there is no workaround for the admin. If you have it as a parameter then admins can use hiera to override it.

spec/acceptance/custom_repo_spec.rb
```diff
@@ -41,7 +41,7 @@ class { 'php::globals':
       it { is_expected.to be_enabled }
     end
     describe command('php --version') do
-      its(:stdout) { is_expected.to match %r{5\.6} }
+      its(:stdout) { is_expected.to match %r{8\.0} }
```
ekohl, line 44: I think [include](https://relishapp.com/rspec/rspec-expectations/v/3-9/docs/built-in-matchers/include-matcher) is a better matcher here.
```suggestion
      its(:stdout) { is_expected.to include('8.0') }
```

spec/acceptance/php_spec.rb
```diff
@@ -14,12 +14,8 @@
       packagename = 'php7.4-fpm'
     when %r{ubuntu-18.04}
       packagename = 'php7.2-fpm'
-    when %r{ubuntu-16.04}
```
ekohl, line 17: If you're removing it here, perhaps it should also be changed on line 39. That can probably be `when %r{ubuntu-}` to catch all Ubuntu versions. Or `if default[:platform].startwith('ubuntu')` instead of a `case` statement.

manifests/repo/redhat.pp
```diff
@@ -22,13 +20,4 @@
     gpgkey     => 'https://rpms.remirepo.net/RPM-GPG-KEY-remi',
     priority   => 1,
   }
-
-  yumrepo { 'remi-php56':
```
ekohl, line 26: Shouldn't this stay with `ensure => absent`?

metadata.json
attachmentgenie: #616

.github/workflows/ci.yml
attachmentgenie: #617

manifests/repo/debian.pp
attachmentgenie: done

manifests/repo/redhat.pp
attachmentgenie: done

attachmentgenie:
getting close, i just need to work out why the 2nd centos 8 puppet runs arent clean

ekohl:
Just a thought. In my experience large PRs tend to be hard to review. In this case, running `php56_spec.rb` is a problem. How about we merge a modulesync that only runs `php_spec.rb` and leave out all the other changes. Then solve those via a separate PR. This allows all other work to continue using Github Actions.

I'm a bit cautious with wiping out all your changes though. What do you think?

attachmentgenie:
> Just a thought. In my experience large PRs tend to be hard to review. In this case, running `php56_spec.rb` is a problem. How about we merge a modulesync that only runs `php_spec.rb` and leave out all the other changes. Then solve those via a separate PR. This allows all other work to continue using Github Actions.
> 
> I'm a bit cautious with wiping out all your changes though. What do you think?

That is a elegant way of getting out of the particular rabbit hole. I agree a simple module sync shouldnt end up being a BC break.
I just followed the code but i must agree with you it all got a bit out of hand. I force pushed the minimum set of changes the get `php_spec` passing on the full matrix and the build is green.  ill make sure to create tickets for the `custom_php_version` cleanup

metadata.json
c33s: don't know what modulesync exactly does, but why does it require a higher puppet version?

manifests/repo/debian.pp
```diff
@@ -2,66 +2,30 @@
...
 #   Public key in apt::key format
 #
-# [*dotdeb*]
-#   Enable special dotdeb handling
-#
-# [*sury*]
-#   Enable special sury handling
-#
 class php::repo::debian (
-  $location     = 'https://packages.dotdeb.org',
```
c33s, line 27: definitly against being able to set a custom repo for debian. removing this is a nogo for me.

manifests/repo/debian.pp
```diff
@@ -2,66 +2,30 @@
...
   assert_private()
 
   include 'apt'
 
-  apt::source { "source_php_${release}":
-    location => $location,
-    release  => $release,
-    repos    => $repos,
+  apt::source { 'source_php_sury':
+    location => 'https://packages.sury.org/php/',
```
c33s, line 23: quite the same as the comment above, but this should be customizable or only drop-in if the requested php version cannot be fullfilled by the debian repo

spec/acceptance/php_spec.rb
c33s: it simply should not be removed. why removing stuff which is working? there is still ubuntu security support for that version. https://en.wikipedia.org/wiki/Ubuntu#Releases

c33s (changes requested):
in gerneral why does a modulesync PR changes so many other parameters (such as default php version set to 8) and is introduces BC breaks?
this should PR should be split into different PR which can be reviewed and discussed separately.

metadata.json
attachmentgenie: it does not perse,

 but it was easier to get this in while fixing the test matrix. Than to fix the matrix and than immediately after remove the support for things like puppet 5 and centos 6/

spec/acceptance/php_spec.rb
attachmentgenie: removing 16.04 is perhaps a bit premature, but the EOL date is in 6 weeks so i removed it already to be a bit efficient

manifests/repo/debian.pp
attachmentgenie: fair point i ll rework

manifests/repo/debian.pp
attachmentgenie: fair point i ll rework

manifests/repo/debian.pp
attachmentgenie: done

manifests/repo/debian.pp
attachmentgenie: done

ekohl (approved):
I've removed all other changes from the PR so it can be merged. Then we can have separate PRs for the unrelated changes.

ekohl:
And we should merge with CentOS 6 acceptance tests failing. Those are not too concerning.

PR #611, opened 2020-12-01
(#607) add 'listen.acl_users' to defined type

### LINKED ISSUE #607: Setting 'listen.acl_users' isn't supported yet

- Module version: latest (bced0bf9f771dc1513d86f3b85692e386c35ced6)

## What are you seeing

`listen.acl_users` is not present in `templates/fpm/pool.conf.erb` yet

Being able to configure `listen_acl_users` like `listen_owner` or `listen_group`

kBite:
enable configuring 'listen.acl_users' by adding that option to pool template

Fixes #607


manifests/fpm/pool.pp
```diff
@@ -24,6 +24,8 @@
 #
 # [*listen_mode*]
 #
+# [*listen_acl_users*]
+#
```
bastelfreak, line 28: please add a short description for the new parameter

manifests/fpm/pool.pp
```diff
@@ -128,6 +130,7 @@
   $listen_owner                            = undef,
   $listen_group                            = undef,
   $listen_mode                             = undef,
+  $listen_acl_users                        = undef,
```
bastelfreak, line 133: can you please add an appropriate datatype? I'm not sure about all allowed values, but I guess `Optional[String[1]]` is a good start?

c33s: https://www.php.net/manual/en/install.fpm.configuration.php

> When POSIX Access Control Lists are supported you can set them using this option. When set, listen.owner and listen.group are ignored. Value is a comma separated list of user names. Since PHP 5.6.5. 

as the values are comma separated it would maybe make sense to use an array of strings.
if we add acl_users it maybe also  would make sense to add `listen.acl_groups` too

PR #620, opened 2021-04-16
puppet/archive: allow 5.x

bastelfreak:


foxxx0 (approved):

bastelfreak:
CentOS 6 fails because it's EoL and their mirrors are dead

PR #621, opened 2021-04-28
Change datatype of `listen.?` parameters from `String[1]` to `Array[String[1]]`

kBite:
Follow-up on #611 implementing `listen.?` parameters as arrays instead of strings.


manifests/fpm/pool.pp
```diff
@@ -121,50 +129,52 @@
...
-  $root_group                              = $php::params::root_group,
-  Optional[Stdlib::Absolutepath] $base_dir = undef,
+  $ensure                                       = 'present',
+  $listen                                       = '127.0.0.1:9000',
+  $listen_backlog                               = '-1',
+  $listen_allowed_clients                       = undef,
+  $listen_owner                                 = undef,
+  $listen_group                                 = undef,
+  $listen_mode                                  = undef,
+  Optional[Array[String[1]]] $listen_acl_users  = undef,
```
bastelfreak, line 139: instead of defaulting to Optional, can you please do this:
```suggestion
  Array[String[1]] $listen_acl_users  = [],
```

this is also documented in our style guide at https://voxpupuli.org/docs/reviewing_pr/

kBite: Than this
```
  # 'php-fpm' expects a comma separated list of user names
  if $listen_acl_users {
    $real_listen_acl_users = join(flatten($listen_acl_users).unique, ",")
  }
```
should become that
```
  # 'php-fpm' expects a comma separated list of user names
  unless $listen_acl_users.empty {
    $real_listen_acl_users = join(flatten($listen_acl_users).unique, ",")
  }
```

Right?

kenyon: That would work, or you can remove the conditional:
```puppet
$real_listen_acl_users = join(flatten($listen_acl_users).unique, ",")
```
Then change the template to
```ruby
if !@real_listen_acl_users.empty?
```

kBite: @kenyon What would be the prefered way?

kenyon: Not sure it matters really. Either way.

kBite: I opted for `unless $listen_acl_users.empty { }`

manifests/fpm/pool.pp
```diff
@@ -186,6 +196,16 @@
     default   => $php::fpm::package,
   }
 
+  # 'php-fpm' expects a comma separated list of user names
+  unless $listen_acl_users.empty {
+    $real_listen_acl_users = join(flatten($listen_acl_users).unique, ",")
```
bastelfreak, line 201:
```suggestion
    $real_listen_acl_users = join(flatten($listen_acl_users).unique, ',')
```

manifests/fpm/pool.pp
```diff
@@ -186,6 +196,16 @@
...
+  }
+
+  # 'php-fpm' expects a comma separated list of group names
+  unless $listen_acl_groups.empty {
+    $real_listen_acl_groups = join(flatten($listen_acl_groups).unique, ",")
```
bastelfreak, line 206:
```suggestion
    $real_listen_acl_groups = join(flatten($listen_acl_groups).unique, ',')
```

bastelfreak: otherwise puppet-lint complains :)

PR #634, opened 2021-08-19
Add/Fix data types of all parameters

smortex:

~~No functional change.  Fixes for invalid/wrong/weird data types will be part of another PR.~~

The initial intention was to have a backward compatible PR (this one) and one that fixes the wrong types, but because the code is inconsistent, it is not possible to add data types in a backward compatible way.

n/a


manifests/params.pp
```diff
@@ -20,12 +20,12 @@
...
-      'pm_max_spare_servers'      => 35,
-      'pm_min_spare_servers'      => 5,
-      'pm_start_servers'          => 5,
-      'request_terminate_timeout' => 0,
+      'pm_max_children'           => '50',
+      'pm_max_requests'           => '0',
+      'pm_max_spare_servers'      => '35',
+      'pm_min_spare_servers'      => '5',
+      'pm_start_servers'          => '5',
+      'request_terminate_timeout' => '0',
```
smortex, line 28: This change is only for consistency with the default data type used in `manifests/fpm/pool.pp` which is `String`.

smortex: This is fixed a few commits later in the PR.

smortex:
So, I do not know how much we can trust the test suite :clown_face:… To be sure I applied the changes on our production infrastructure and it seems to be fine :smile: (I only saw changes from PR which have already been merged into the `master` branch).

Ready for review!

PR #636, opened 2021-08-26
Drop Debian 9/Ubuntu 16.04 support

root-expert:


thatgraemeguy:
I realise this is too little too late for this instance, but for future reference perhaps its worth knowing that Debian does have an LTS support phase that lasts an additional 2 years beyond the end of the main 5-year support.

bastelfreak:
we're aware of that, but we do not support the extended support lifecycles that Debian/Ubuntu offer. We try to support stable+oldstable which is already challenging enough.

thatgraemeguy:
OK, thanks for clarifying.

PR #642, opened 2021-09-05
Allow 0 for pm_start_servers

jkroepke:
pm_start_servers=0 is a supported setting, can be used for on demand services.


smortex (approved):
LGTM

smortex:
8.0.2 was released with the fix.  Thanks!

PR #645, opened 2021-09-17
Fix static redhat remi repo v8.0.2

kobybr:
Currently the class php::repo::redhat is hardcoded to create only the remi56 repo.   This request removes that static value and creates based on $yum_repo.   This allows for other versions such as remi_php72, etc..

### DIFF
--- a/manifests/repo/redhat.pp
+++ b/manifests/repo/redhat.pp
@@ -25,7 +25,7 @@ class php::repo::redhat (
 
   yumrepo { $yum_repo:
     descr      => "Remi's ${yum_repo.match('php.+$')[0].strip} RPM repository for Enterprise Linux \$releasever - \$basearch",
-    mirrorlist => "https://rpms.remirepo.net/enterprise/$releasever/${yum_repo.match('php.+$')[0].strip}/mirror",
+    mirrorlist => "https://rpms.remirepo.net/enterprise/${releasever}/${yum_repo.match('php.+$')[0].strip}/mirror",
     enabled    => 1,
     gpgcheck   => 1,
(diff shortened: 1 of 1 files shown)


manifests/repo/redhat.pp
```diff
@@ -23,9 +23,9 @@
     priority   => 1,
   }
 
-  yumrepo { 'remi-php56':
```
smortex, line 26: Might be better to cleanup if a version is changed by using a fixed-name repo and removing the old one, e.g.:

```puppet
yumrepo { 'remi-php56':
  ensure => absent,
}

yumrepo { 'remi-php':
  ...
}
```

manifests/repo/redhat.pp
```diff
@@ -23,9 +23,9 @@
...
-    descr      => 'Remi\'s PHP 5.6 RPM repository for Enterprise Linux $releasever - $basearch',
-    mirrorlist => "https://rpms.remirepo.net/enterprise/${releasever}/php56/mirror",
+  yumrepo { $yum_repo:
+    descr      => "Remi's ${yum_repo.match('php.+$')[0].strip} RPM repository for Enterprise Linux \$releasever - \$basearch",
+    mirrorlist => "https://rpms.remirepo.net/enterprise/$releasever/${yum_repo.match('php.+$')[0].strip}/mirror",
```
smortex, line 28: Missing {} break CI:
```suggestion
    mirrorlist => "https://rpms.remirepo.net/enterprise/${releasever}/${yum_repo.match('php.+$')[0].strip}/mirror",
```

smortex (changes requested):
:wave: Hi!

I added a few in-line notes.

The test will probably need to be updated to take these changes into account, and extended to make sue we do not break this in the future.

Thanks!

PR #650, opened 2022-01-05
Allow environment variables when configuring FPM pools

### LINKED ISSUE #649: Data type validation in fpm::pool breaks ability to use environment variables as values

- Puppet: Any
- Ruby: Any
- Distribution: Any
- Module version: 8.0+

```
php::fpm::pool { 'www':
  ..
  pm_max_children => '${FPM_MAX_CHILDREN}',
}
```

## What are you seeing

The validation added in #634 appears to have broken the ability to configure the PM settings via environment variables (interpreted at runtime), which is something we do to allow our environment to determine the PHP runtime settings.

Per the PHP docs, this is a supported configuration: https://www.php.net/manual/en/configuration.file.php#example-1

Suggestion would be to make these either `Variant[String, Integer[..]]` or for a more strict approach something like, `Variant[Pattern[/\$\{[\w]+\}/], Integer[..]]`.

No Puppet errors, environment variable based configuration written out to file.

## Output log

```
Error: Evaluation Error: Error while evaluating a Resource Statement, Php::Fpm::Pool[www]:
  parameter 'pm_max_children' expects an Integer value, got String
  parameter 'pm_start_servers' expects an Integer value, got String
  parameter 'pm_min_spare_servers' expects an Integer value, got String
  parameter 'pm_max_spare_servers' expects an Integer value, got String
  parameter 'pm_max_requests' expects an Integer value, got String
```

chrisboulton:

Allows for PHP FPM pool settings to be configured as environment variables (`pm_max_children => '${PM_MAX_CHILDREN}'`) in addition to their existing typed formats.

I debated about the best way to add this, and ended up adding `Variant` types, along with a single custom type. I did this because there were varying types (`Integer[0], Integer[1], Php::Duration`) that I didn't want to have to duplicate.

The env based configuration applies to every setting, but for right now I've added it to the most common ones where due to the stricter validation, it would previously fail.

Fixes #649


smortex (commented):
Maybe some dumb reflexion since I do not use PHP much…

My feeling is that any parameter which does not accept `String` should accept a `Php::EnvString`, right?  I see some `Stdlib::Absolutepath`, `Stdlib::Filemode`, `Integer`, `Enum['yes', 'no']`: should these also be adjusted to allow variables?

types/envstring.pp
```diff
@@ -0,0 +1 @@
+type Php::EnvString = Pattern[/\$\{[\w]+\}/]
```
smortex, line 1: The square brackets look redundant and I think we should anchor the regexp:
```suggestion
type Php::EnvString = Pattern[/\A\$\{\w+\}\z/]
```

yakatz:
@chrisboulton If you are still interested and can rebase this, that would be great. I might have time in the next few days otherwise.

Issue #653, opened 2022-01-13
Unable to remove extension configuration settings


mpdude:
When using this: 

```puppet
::php::extension { 'some-extension':
    settings => {
        'some.setting' => 'old-value',
    }
}
```

... and you later change it to ...

```puppet
::php::extension { 'some-extension':
    settings => {}
}
```

... the `some.setting` is not removed from the `...some-extension.ini` file. 

The reason probably is that the `settings` hash is converted into `::php::config::setting`s here:

https://github.com/voxpupuli/puppet-php/blob/82a1698d05d6e2f8859ad506a66095f580a68c1a/manifests/config.pp#L28

... which in turn becomes an `ini_setting`:

https://github.com/voxpupuli/puppet-php/blob/82a1698d05d6e2f8859ad506a66095f580a68c1a/manifests/config/setting.pp#L43-L49

... but there is never anything that cleans up old/stale/unmanaged settings.

In an older version of this module, a possible workaround was to use `undef` for values that you'd like to remove, but that's no longer possible due to the typing. #654 re-enables that.

## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet:
- Ruby:
- Distribution: Ubuntu 20.04
- Module version: 8.0.2

PR #654, opened 2022-01-13
Allow ::php::config::setting value to be "undef"

mpdude:
This is necessary so that config settings can be removed (at least as a workaround for #653).

### DIFF
--- a/manifests/config/setting.pp
+++ b/manifests/config/setting.pp
@@ -20,7 +20,7 @@
 #
 define php::config::setting (
   String[1] $key,
-  Optional[Variant[Integer, String]] $value,
+  Variant[Integer, String, Undef] $value,
   Stdlib::Absolutepath $file,
 ) {
   assert_private()
--- a/manifests/config/setting.pp
+++ b/manifests/config/setting.pp
@@ -20,7 +20,7 @@
 #
 define php::config::setting (
   String[1] $key,
-  Variant[Integer, String] $value,
+  Optional[Variant[Integer, String]] $value,
   Stdlib::Absolutepath $file,
 ) {
   assert_private()


mpdude:
I think I need some help or starting pointers regarding tests... :-(

smortex:
`php::config::setting` is private so cannot be tested directly.  You can however add tests for `php::config`: they live in `spec/defines/config_spec.rb`.  Here is an example around existing code, feel free to add more test cases:

```diff
diff --git a/spec/defines/config_spec.rb b/spec/defines/config_spec.rb
index 77976bb..5cc15e1 100644
--- a/spec/defines/config_spec.rb
+++ b/spec/defines/config_spec.rb
@@ -83,6 +83,15 @@ describe 'php::config' do
           end
 
           it { is_expected.to contain_php__config('unique-name').with_file('/etc/php5/conf.d/unique-name.ini') }
+          it do
+            is_expected.to contain_ini_setting('/etc/php5/conf.d/unique-name.ini: apc.enabled').with(
+              'ensure'  => 'present',
+              'path'    => '/etc/php5/conf.d/unique-name.ini',
+              'section' => '',
+              'setting' => 'apc.enabled',
+              'value'   => 1,
+            )
+          end
         end
 
         context 'empty array' do
```

mpdude:
@smortex Thank you!

I have looked at that file, but cannot find another example of how to initialize (stage; setup fixture) pre-exisiting config file contents.

Also, what is `contain_ini_setting` from your diff? I cannot find that being used elsewhere.

smortex:
The example in my diff above is I think the kind of test that is missing: your new parameter allows to manage the `ensure` parameter of the `ini_setting` resource, and a test should me added to test this.  But since there are currently no tests for the `ini_setting` resource, I added one for the first test case to show how it works and so you can write more.

After applying this diff, you can run the test suite with:

```
bundle install # if you have not already done this
bundle exec rake spec
```

Tests are performed with https://rspec-puppet.com/ and `contain_ini_setting()` is used to test for a `ini_setting` resource, see: https://rspec-puppet.com/documentation/classes/#test-a-resource

Does it help?

root-expert:
Was implemented at #647

PR #655, opened 2022-02-22
fix installing latest PHP on Debian

### LINKED ISSUE #458: Why does dotdeb APT wheezy 5.6 repo get installed on Debian 9?
Hello,

I am using the following hiera node config in order to install PHP 7.1 from the sury repo on a Debian 9 server:

```
php::globals::php_version: '7.1'
php::manage_repos: true
php::repo::debian::dotdeb: false
```

and noticed that this also installs the dotdeb APT repo for wheeezy PHP 5.6 as you can see in the `/etc/apt/sources.list.d/source_php_wheezy-php56.list` file:

```
# This file is managed by Puppet. DO NOT EDIT.
# source_php_wheezy-php56
deb http://packages.dotdeb.org wheezy-php56 all
```

Is this required? or is it a bug maybe?

I am using puppet 4.8.2 on Debian 9 servers and the latest version of this forge module (v.5.3.0).

Thanks

bwana147:
I tried installing PHP 7.4 on Debian bullseye, and ended up with package from dotdeb, with references to php 5.6, on Debian wheezy. Some parameters to the php::repo::debian class were just ignored, and the test on php_version being 7.1 or 7.2 seemed too restrictive to me. So I fixed the code so that it works for me, but please keep in mind that I'm by no means a Puppet expert, and even less so when it comes to PHP.
And this is my first Pull Request, so I hope I did not mess up.

Fixes #458

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -35,22 +35,22 @@ class php::repo::debian (
   Boolean $dotdeb         = true,
   Boolean $sury           = true,
 ) {
-  assert_private()
-
   include 'apt'
 
-  apt::source { "source_php_${release}":
-    location => $location,
-    release  => $release,
-    repos    => $repos,
-    include  => {
-      'src' => $include_src,
-      'deb' => true,
-    },
-    key      => $key,
+  if ($dotdeb) {
+    apt::source { "source_php_${release}":
+      location => $location,
+      release  => $release,
+      repos    => $repos,
+      include  => {
+        'src' => $include_src,
+        'deb' => true,
+      },
+      key      => $key,
+    }
   }
 
-  if ($sury and $php::globals::php_version in ['7.1','7.2']) {
+  if ($sury) {
     apt::source { 'source_php_sury':
       location => 'https://packages.sury.org/php/',
       repos    => 'main',
--- a/spec/classes/php_repo_debian_spec.rb
+++ b/spec/classes/php_repo_debian_spec.rb
@@ -1,0 +1,29 @@
+require 'spec_helper'
+
+describe 'php::repo::debian', type: :class do
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let :facts do
+        facts
+      end
+
+      describe 'works without params' do
+        if facts[:os]['name'] == 'Debian'
+          it { is_expected.to compile.with_all_deps }
+          context 'dotdeb' do
+            let(:params) { { 'dotdeb' => true, 'sury' => false } }
+
+            it { is_expected.to contain_apt__source('source_php_wheezy-php56') }
+            it { is_expected.not_to contain_apt__source('source_php_sury') }
+          end
+          context 'sury' do
+            let(:params) { { 'dotdeb' => false, 'sury' => true } }
+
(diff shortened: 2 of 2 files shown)


smortex:
Can you please remove commit 13b1f3d (voxpupuli does not use the PDK)?

From your working directory:

```
git rebase -i origin/master # Interactively rewrite history
```

This will bring your editor listing all commits in your branch.  Remove the line corresponding to commit 13b1f3d and exit.
Then push your updated branch to update this PR:

```
git push -f            # Send the changes (-f is required because we re-wrote history)
```

bwana147:
Thank you for your reply. I did as you said, then there is some CI process that complained about code not complying with coding standards, so I git pushed --force again... HTH

manifests/repo/debian.pp
```diff
@@ -35,22 +35,22 @@
   Boolean $dotdeb         = true,
   Boolean $sury           = true,
 ) {
-  assert_private()
```
root-expert, line 38: This class was supposed to be private. any reason this was changed?

manifests/repo/debian.pp
```diff
@@ -35,22 +35,22 @@
...
+      repos    => $repos,
+      include  => {
+        'src' => $include_src,
+        'deb' => true,
+      },
+      key      => $key,
+    }
   }
 
-  if ($sury and $php::globals::php_version in ['7.1','7.2']) {
```
root-expert, line 53: I believe here we should add 7.4 (and possible 7.3) instead if removing it.

manifests/repo/debian.pp
bwana147: Well, I'm not really savvy with puppet... So I converted the module to pdk (since I felt it made it simpler for me by hiding the ruby/gem/rake/whatever that lies behind). And then `pdk test unit --tests=./spec/classes/php_repo_debian_spec.rb` would not work. Maybe it's because my environment is poorly set up? How am I supposed to run the tests without pdk?

manifests/repo/debian.pp
bwana147: I thought it would be less troublesome to assume that the requested version was supported by Sury, rather than keeping an eye on them and patch puppet-php each time they release a new version (they must be shipping 8.0, or maybe even 8.1).

manifests/repo/debian.pp
smortex: > How am I supposed to run the tests without pdk?

```
bundle exec rake spec
```

This assume you have a bundle, if not run something like `bundle install --path /tmp/gems` to install gems in a scatch directory without needing to be root

root-expert: Take a look at [CONTRIBUTING](https://github.com/voxpupuli/puppet-php/blob/master/.github/CONTRIBUTING.md) for more info as well

bwana147:
Sorry guys... I thought that I'd be able to help improving this module but it looks like I'm not. As I said, I am no Ruby developper and have no time to get acquainted with the whole Ruby ecosystem (gem, bundle, rake, beaker, whatever... They are just words to me, and it looks like a lot of time to learn them all). So I won't be able to produce a proper commit for this issue. Thank you anyway for your time.

PR #656, opened 2022-03-09
Composer

kapouik:
By default, when the module install composer, it use https://getcomposer.org/composer.phar
It install the latest preview from nighty build of the dev main repository.

I have set a case that allow to download the right version of composer depending from the channel selected. By default, it use the latest stable.

No issue


manifests/composer.pp
```diff
@@ -38,9 +38,31 @@
...
+      $main_source = 'https://getcomposer.org/download/latest-1.x/composer.phar'
+    }
+    '2': {
+      $main_source = 'https://getcomposer.org/download/latest-2.x/composer.phar'
+    }
+    default: {
+      fail('Channel not supported by composer')
+    }
+  }
+  $final_source = pick( $source, $main_source )
```
smortex, line 61: How come puppet-lint not complain with space inside parenthesis? :shrug: 

```suggestion
  $final_source = pick($source, $main_source)
```

manifests/composer.pp
```diff
@@ -27,7 +27,7 @@
 #   UNIX group of the root user
 #
 class php::composer (
-  String $source                       = $php::params::composer_source,
+  Optional[String[1]] $source          = undef,
```
smortex, line 30: This seems to be a fixed list.  Does it make sense to change it to `Optional[Enum['stable', 'preview', ...]]`?

manifests/composer.pp
```diff
@@ -38,9 +38,31 @@
 ) inherits php::params {
   assert_private()
 
+  case $channel {
```
smortex, line 41: Usage of a selector would be more concise:

```puppet
$main_source = $channel ? {
  stable  => '...',
  preview => '...',
  # ...
}
```

smortex (commented):
Apart from the spacing comment which does not match the rest of voxpupuli code, the rest is under your appreciation :wink:

manifests/composer.pp
kapouik: $source contain an URL. may be better to find a type URL (must look on it).
this var may be usefull to override the default selector for a local mirror.

manifests/composer.pp
c33s: my thoughts would be:

the `$channel` variable is a custom type (basically Enum), so we could skip the error handling, remove the case statement and instead use a hash.

i would add a new variable called `$channel_sources` and put the predefined channel sources in  as hash. so channel sources can easily be overwritten if required.

https://github.com/voxpupuli/puppet-php/blob/master/data/default.yaml
```yaml
channel_sources:
  stable:   'https://getcomposer.org/download/latest-stable/composer.phar'
  preview:  'https://getcomposer.org/download/latest-preview/composer.phar'
  snapshot: 'https://getcomposer.org/composer.phar'
  1:        'https://getcomposer.org/download/latest-1.x/composer.phar'
  2:        'https://getcomposer.org/download/latest-2.x/composer.phar'
```

manifests/composer.pp
smortex: D'oh! I mixed up *source* and *channel*, ignore this enum thing.

> $source contain an URL. may be better to find a type URL (must look on it).

[Stdlib::HTTPUrl](https://github.com/puppetlabs/puppetlabs-stdlib#stdlibhttpurl) :wink:

manifests/composer.pp
smortex: Are some URL OS-dependant?  If not, I am not sure the ability to override them in hiera is needed: the `$source` parameter already allows to set a custom value and the module alway provide the well-known values regardless of what people put in hiera.

c33s: to override something can always be helpful. for example this or another module (can't remember) had the apt repo key hardcoded like this. this is not a problem on high active modules where PR are merged fast but on repos where fixes are only merged very slow it requires you to fork the repo and fix the key yourself. to allow the user to override static values is, in my perspective, a best practice.

a usecase could be that i have an internal mirror of all composer channels, with static config in hiera i simply can choose my own channelurls without touch the modules code.

manifests/composer.pp
kapouik: This code no more exist. I have transform it in a better form.

manifests/composer.pp
kapouik: I have made the modification suggested by @c33s. I have made some tests on a real puppet installation and the result is correct.
If it's ok, I can made a rebase to merge all my commit into one.

manifests/composer.pp
```diff
@@ -27,7 +27,7 @@
 #   UNIX group of the root user
 #
 class php::composer (
-  String $source                       = $php::params::composer_source,
+  Hash $channel_sources                = undef,
```
root-expert, line 30:
```suggestion
  Hash $channel_sources                = {},
```

manifests/composer.pp
```diff
@@ -40,7 +40,7 @@
 
   archive { 'download composer':
     path         => $path,
-    source       => $source,
+    source       => $channel_sources[$channel],
```
root-expert, line 43: I would use dig here since it's undef-safe

```suggestion
    source       => $channel_sources.dig($channel),
```

root-expert (changes requested):
Hello @kapouik ,

can we get some basic unit tests for this change?

kapouik:
I have prepare basic test : It only check if composer is installed from default fact. 

As other test, I can add check if other channel work if I change it. If you have other idea ?

yakatz:
This PR has had no activity in a long time. Feel free to rebase/update/resubmit.

PR #663, opened 2022-07-02
Arch Linux: Fix package names and set correct php-fpm user (root -> http)

jkroepke:

PR #664, opened 2022-07-03
php-fpm: Reload service with systemd if available

bastelfreak:
newer distributions, like Arch Linux, do not ship `service` anymore. If systemd is the used init system, we should also use the systemd cli to reload a service. This is required to fix Arch Linux acceptance tests (besides #663 )


jkroepke:
Question: Why not trigger the service resource? Is there a real benefit of manually execute the reload command?

bastelfreak:
I guess people wanted to avoid a restart and do and explicit reload.

PR #669, opened 2022-10-05
Support other versions of remi repo on redhat

kobybr:
Currently the class php::repo::redhat is hardcoded to create only the remi56 repo. This request removes that static value and creates based on $yum_repo. This allows for other versions such as remi_php72, etc..

### DIFF
--- a/manifests/repo/redhat.pp
+++ b/manifests/repo/redhat.pp
@@ -7,12 +7,14 @@
 #
 
 class php::repo::redhat (
-  String[1] $yum_repo = 'remi_php56',
+  $version = '5.6',
 ) {
+
   $releasever = $facts['os']['name'] ? {
     /(?i:Amazon)/ => '6',
     default       => '$releasever',  # Yum var
   }
+  $no_dot_version = $version.delete('.')
 
   yumrepo { 'remi':
     descr      => 'Remi\'s RPM repository for Enterprise Linux $releasever - $basearch',
@@ -23,9 +25,9 @@ class php::repo::redhat (
     priority   => 1,
   }
 
-  yumrepo { 'remi-php56':
-    descr      => 'Remi\'s PHP 5.6 RPM repository for Enterprise Linux $releasever - $basearch',
-    mirrorlist => "https://rpms.remirepo.net/enterprise/${releasever}/php56/mirror",
+  yumrepo { "remi-php${no_dot_version}":
+    descr      => "Remi's PHP ${version} RPM repository for Enterprise Linux \$releasever - \$basearch",
+    mirrorlist => "https://rpms.remirepo.net/enterprise/${releasever}/php${no_dot_version}/mirror",
     enabled    => 1,
     gpgcheck   => 1,
     gpgkey     => 'https://rpms.remirepo.net/RPM-GPG-KEY-remi',
--- a/manifests/repo/redhat.pp
+++ b/manifests/repo/redhat.pp
@@ -7,9 +7,8 @@
 #
 
 class php::repo::redhat (
-  $version = '5.6',
+  String[1] $version = '5.6',
 ) {
-
   $releasever = $facts['os']['name'] ? {
     /(?i:Amazon)/ => '6',
     default       => '$releasever',  # Yum var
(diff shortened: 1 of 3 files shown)


manifests/repo/redhat.pp
```diff
@@ -7,12 +7,13 @@
 #
 
 class php::repo::redhat (
-  String[1] $yum_repo = 'remi_php56',
+  String[1] $version = '5.6',
```
smortex, line 10: Wondering if we should use an explicit list of allowed values (`Enum['5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2']`) which makes it obvious when reading the doc what is allowed here but needs to be updated quite often, or match something like `Pattern['\A\d+\.\d+\z']`?

While speaking about the doc strings, pleas update them in the top of the file.

manifests/repo/redhat.pp
```diff
@@ -23,9 +24,9 @@
     priority   => 1,
   }
 
-  yumrepo { 'remi-php56':
-    descr      => 'Remi\'s PHP 5.6 RPM repository for Enterprise Linux $releasever - $basearch',
-    mirrorlist => "https://rpms.remirepo.net/enterprise/${releasever}/php56/mirror",
+  yumrepo { "remi-php${no_dot_version}":
```
smortex, line 27: There might be leftovers when we change `$version`, is it safe or can it be an issue?

smortex (changes requested):
Nice, I added some in-line comments for some points of attention and the docstrings need updating.  Thanks!

manifests/repo/redhat.pp
```diff
@@ -7,12 +7,13 @@
 #
 
 class php::repo::redhat (
-  String[1] $yum_repo = 'remi_php56',
```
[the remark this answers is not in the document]
kobybr: I can rework the code to keep the original param name if that makes more sense to keep it backwards compatible.

manifests/repo/redhat.pp
kobybr: Sure, I can add that

manifests/repo/redhat.pp
kobybr: It is hard to say.

manifests/repo/redhat.pp
kobybr: I think if anything, the Pattern would make the most sense

manifests/repo/redhat.pp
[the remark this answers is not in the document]
smortex: Only keep it unchanged if it makes sense: breaking backwards compatibility is not something that should stop you :wink:   We (voxpupuli team) just want not to forget about this kind of things so that end-users can trust our minor updates to not break their setup!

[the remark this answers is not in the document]
kobybr: I think it would be better to keep the backwards compatibility - I wasn't really thinking in those terms, and there really isn't a need to change it other than cosmetics.   I pushed the change.

manifests/repo/redhat.pp
kobybr: I'm not really sure how to go about cleanup if a version is changed.   I would just assume leave it up to the end-user to do an ensure absent in their control repo.   Open to suggestions though.

kobybr:
Just checking in on this.  Seems I have all the pieces in place, let me know otherwise.

Thanks

farlerac:
Just checking in, are there any more issues we need to address with this PR?

manifests/repo/redhat.pp
```diff
@@ -13,6 +13,7 @@
     /(?i:Amazon)/ => '6',
     default       => '$releasever',  # Yum var
   }
+  $version = $yum_repo.match('.\d$')[0]
```
smortex, line 16: `.\d$` feels weird: any char followed by a digit at the end of a line.

I guess what you mean here is any number of digits at the end of the string? `\d+\z`

smortex (changes requested):
Oops, I missed the notification. Just a minor hitch.

PR #674, opened 2022-12-13
Add option to disable management of rundir

tmanninger:
I am using php7.4 from sury on debian 10.
After every server reboot, this module creates the `/var/run/php-fpm` directory (which is not used by php-fpm) and triggers a reload, which resets all open connections to the php-fpm daemon.

This PR added an option to disable the managing of the `/var/run/php-fpm` directory

### DIFF
--- a/manifests/fpm/config.pp
+++ b/manifests/fpm/config.pp
@@ -70,6 +70,9 @@
 # [*pid_file*]
 #   Path to fpm pid file
 #
+# [*manage_run_dir*]
+#   Manage the run directory
+#
 class php::fpm::config (
   Stdlib::Absolutepath $config_file                                     = $php::params::fpm_config_file,
   String $user                                                          = $php::params::fpm_user,
@@ -93,6 +96,7 @@ class php::fpm::config (
   String[1] $root_group                                                 = $php::params::root_group,
   String $syslog_facility                                               = 'daemon',
   String $syslog_ident                                                  = 'php-fpm',
+  Boolean $manage_run_dir                                               = true
 ) inherits php::params {
   assert_private()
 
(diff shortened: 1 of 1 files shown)


kenyon:
Why are we managing this directory if it's not used? Maybe there is a better solution here, like making it conditional on some version of package or OS?

tmanninger:
We are using /var/run/ and i wouldn't manage this directory with the puppet-php module.

tmanninger:
It there no way to merge this MR? 
I am using the PHP package from packages.sury.org, and i don't know, which OS versions and PHP versions are affected.
I think the best way is it to make it configureable.

kenyon (approved):
Seems OK to me. We'll want to rebase when merging this to eliminate the merge commit.

tmanninger:
@kenyon i removed the merge commit

manifests/fpm/config.pp
```diff
@@ -93,6 +93,7 @@
   String[1] $root_group                                                 = $php::params::root_group,
   String $syslog_facility                                               = 'daemon',
   String $syslog_ident                                                  = 'php-fpm',
+  Boolean $manage_run_dir                                               = true
```
smortex, line 96: Please add documentation for this new parameter in the comment block above. That way it will be added to REFERENCE.md on the next release.

tmanninger:
@smortex i added the documentation

tmanninger:
@smortex is everything fine?

smortex (approved):
This feel weird, but I don't know what this directory is used for and the default behavior is unchanged so LGTM :shrug:.

smortex:
CI failure seems unrelated, merging

PR #688, opened 2023-05-29
fpm: manage runtime- and log directory based on params

UiP9AV6Y:
instead of hardcoding the paths for the log- and runtime directory, calculate them based on the error_log and pid_file params respectively. the directories will not be managed if they are shared system locations like `/var/log` or `/run` (this is checked by ensuring the directories have *php* in their name, signalling that those are paths dedicated to PHP)

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -1119,6 +1119,9 @@ Configure php-fpm service
 [*error_log*]
   Path to error log file. If it's set to "syslog", log is
   sent to syslogd instead of being written in a local file.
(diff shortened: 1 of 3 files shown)


yakatz:
I feel like checking the path name for `php` is going to confuse some users. The code looks fine, but I am not sure this module should be doing this. If you need more complicated path management, you should probably do that in your own manifest.

yakatz:
The consensus (Slack chat) was that this type of change should be done in your own module.

PR #695, opened 2023-08-08
Release 9.0.0

rwaffen:


rwaffen:
since January only arch fails again and again ;)

zilchms (approved):
LGTM, but at some point we should look into those archlinux failures and fix them :p

PR #699, opened 2023-09-21
switch to stdlib namespaced functions, require stdlib 9.x

tuxmea:


manifests/dev.pp
```diff
@@ -33,7 +33,7 @@
 
   # Default PHP come with xml module and no seperate package for it
   if $facts['os']['name'] == 'Ubuntu' {
-    ensure_packages(["${php::package_prefix}xml"], { ensure  => present, require => $require, })
+    stdlib::ensure_packages(["${php::package_prefix}xml"], { ensure  => present, require => $require, })
```
evgeni, line 36: you have to bump the `puppetlabs/stdlib` requirement to `>= 9` for that

evgeni (changes requested):

manifests/dev.pp
evgeni: and given https://github.com/voxpupuli/puppet-php/pull/691 did not end up red when the upper bound was bumped, I guess this line has no test coverage?

evgeni:
(the tests install `puppetlabs-stdlib (v8.6.0)` which explains why it was not caught during bumping)

evgeni:
@bastelfreak would this be considered a breaking change?

technically it's fixing a bug introduced by https://github.com/voxpupuli/puppet-php/pull/691 which allowed stdlib 9.x without using the new namespaced name (*insert raising fist emoji*), so I'd lean to "no", but requirement bumps always give me headaches.

evgeni:
@tuxmea I guess the following places should also be updated?

https://github.com/voxpupuli/puppet-php/blob/d5c3d665a22129274e4c2ef8eda605a1b985c9ec/manifests/phpunit.pp#L26
https://github.com/voxpupuli/puppet-php/blob/d5c3d665a22129274e4c2ef8eda605a1b985c9ec/manifests/pear.pp#L50

tuxmea:
> @tuxmea I guess the following places should also be updated?
> 
> https://github.com/voxpupuli/puppet-php/blob/d5c3d665a22129274e4c2ef8eda605a1b985c9ec/manifests/phpunit.pp#L26
> 
> https://github.com/voxpupuli/puppet-php/blob/d5c3d665a22129274e4c2ef8eda605a1b985c9ec/manifests/pear.pp#L50

Done

tuxmea:
As ensure_packages is available (and shows a deprecation) I would consider this to not be a breaking change. But I am open to other opinions. @bastelfreak what do you prefer?

metadata.json
```diff
@@ -11,7 +11,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.16.0 < 10.0.0"
+      "version_requirement": ">= 9.0.0 < 10.0.0"
```
smortex, line 14: Changing the lower version *is* a breaking change (e.g. users of version 7.0.0 will have unmet dependencies when updating, and ignoring it would break their catalog as the namespaced function is not available).

smortex (approved):
LGTM!

bastelfreak (approved):

rwaffen (approved):

PR #733, opened 2025-02-05
fix: set proxy environment variables correctly

### LINKED ISSUE #731: Support for auto-update via proxy http
First, what a pain to document this bug!  Sorry for the wall of text.

# What is the problem
Composer is installed correctly, but never get updated when using an http proxy:
```puppet
    class { 'php::composer':
      proxy_type   => 'http',
      proxy_server => 'http://myproxy.example.com:3128',
    }
```

I should define `proxy_type` as `https` for the update to work correctly, but is not really intuitive…
```puppet
    class { 'php::composer':
      proxy_type   => 'https',
      proxy_server => 'http://myproxy.example.com:3128',
    }
```

# Source of the problem
The original author of the `php::composer::auto_update` seems to have mix the usage of `http_proxy` and `https_proxy` environment variables.  
The http/https part is not for the communication to the proxy server, but for the destination URL.  The reference to these environment variables are often badly documented on the internet…

Per the manpage of curl(1):
>       http_proxy [protocol://]<host>[:port]
>              Sets the proxy server to use for HTTP.
>
>       HTTPS_PROXY [protocol://]<host>[:port]
>              Sets the proxy server to use for HTTPS.

Therefore, if only `http_proxy` is defined, a request to `https://getcomposer.org` (HTTPS) will not be done through the proxy, only `http://getcomposer.org` (HTTP) will be done through the proxy.

Both `http_proxy` and `https_proxy` should be defined if both requests to `http://getcomposer.org` and `https://getcomposer.org` want to be redirected through proxy, no matter if the proxy uses https or http as protocol.

If my proxy does not use HTTPS as protocol, environment should be defined as
```
http_proxy=http://myproxy.example.com:3128
https_proxy=http://myproxy.example.com:3128
```

If my proxy uses HTTPS as protocol, environment should be defined as
```
http_proxy=https://mytlsproxy.example.com:3128
https_proxy=https://mytlsproxy.example.com:3128
```

### Archive module correctly defined
`proxy_type` in the `archive` module is used to defined the protocol of the proxy only if the `proxy_server` is not a valid URI

```puppet
          uri = URI(options[:proxy_server])
          uri = URI("#{options[:proxy_type]}://#{options[:proxy_server]}") unless uri.scheme
```
https://github.com/voxpupuli/puppet-archive/blob/63474f6709e4c44b1b4c6852b32335f34df32877/lib/puppet_x/bodeco/util.rb#L78

therefore, the `proxy_type` parameter is correctly used in `php::composer` to define the proxy, only if you do not set `proxy_server` as full URI:
```puppet
  archive { 'download composer':
    path         => $path,
    source       => $source,
    proxy_type   => $proxy_type,
    proxy_server => $proxy_server,
  }
```

So you should be able to call this module as both way:
```puppet
    archive { 'example':
      path         => $path,
      source       => $source,
      proxy_type   => 'http',
      proxy_server => 'myproxy.example.com:3128',
    }

    archive { 'example':
      path         => $path,
      source       => $source,
      proxy_server => 'http://myproxy.example.com:3128',
    }
```

### Auto-update
Auto-update does not use this `proxy_type` parameter the same way.

Here is how it is defined in `php::composer::auto_update`:
```puppet
  if $proxy_type and $proxy_server {
    $env = ['HOME=/root', "${proxy_type}_proxy=${proxy_server}"]
  } else {
    $env = ['HOME=/root']
  }
```

# Solutions

### simple
The simple solution is to not use the `proxy_type` parameter at all.
```puppet
  if $proxy_server {
    $env = [
      'HOME=/root',
      "http_proxy=${proxy_server}",
      "https_proxy=${proxy_server}",
    ]
  } else {
    $env = ['HOME=/root']
  }
```

### clean
The cleanest solution in my opinion would be to detect if `proxy_server` is a valid URI the same way the `archive` module do, and throw an error to the user if not correctly defined:

```puppet
  if $proxy_server {
    if proxy_server =~ /^https?:\/\// {
      $_proxy_server = $proxy_server
    } else {
      if $proxy_type {
        $_proxy_server = "${proxy_type}://${proxy_server}"
      } else {
        fail('proxy_type must be defined if proxy_server is not full URI (https://example.com)')
      }
    }
    $env = [
      'HOME=/root',
      "http_proxy=${_proxy_server}",
      "https_proxy=${_proxy_server}",
    ]
  } else {
    $env = ['HOME=/root']
  }
```

I will prepare a PR on this tomorrow, any feedback would be appreciated.

teluq-pbrideau:
Fix the environment variable as documented in curl(1)
Throw an error at the user if not correctly configured

See issue for the full explanation

Fixes #731

### DIFF
--- a/manifests/composer/auto_update.pp
+++ b/manifests/composer/auto_update.pp
@@ -37,9 +37,21 @@ class php::composer::auto_update (
   Optional[String[1]] $proxy_server = undef,
 ) {
   assert_private()
-
-  if $proxy_type and $proxy_server {
-    $env = ['HOME=/root', "${proxy_type}_proxy=${proxy_server}"]
+  if $proxy_server {
+    if $proxy_server =~ Stdlib::HTTPUrl {
+      $_proxy_server = $proxy_server
+    } else {
+      if $proxy_type {
+        $_proxy_server = "${proxy_type}://${proxy_server}"
+      } else {
+        fail('proxy_type must be defined if proxy_server is not full URL (https://example.com)')
+      }
+    }
+    $env = [
+      'HOME=/root',
+      "http_proxy=${_proxy_server}",
+      "https_proxy=${_proxy_server}",
+    ]
   } else {
     $env = ['HOME=/root']
   }
--- a/spec/classes/php_composer_spec.rb
+++ b/spec/classes/php_composer_spec.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'php::composer', type: :class do
+  on_supported_os.each do |os, facts|
+    context "on #{os}" do
+      let :facts do
+        facts
+      end
+
+      let :pre_condition do
+        'class { "php": composer => false }'
+      end
+
+      describe 'works without params' do
+        it { is_expected.to compile.with_all_deps }
+      end
+
+      describe 'when called with no parameters' do
+        it { is_expected.to contain_class('php::composer::auto_update') }
+        it { is_expected.to contain_archive('download composer') }
+      end
+
+      describe 'when auto_update is disabled' do
+        let(:params) { { auto_update: false } }
+
+        it { is_expected.not_to contain_class('php::composer::auto_update') }
+      end
+
+      describe 'when called with proxy_server parameters' do
+        let(:params) { { proxy_server: 'http://example.com' } }
+
+        it { is_expected.to contain_archive('download composer').with(proxy_server: 'http://example.com') }
+        it { is_expected.to contain_exec('update composer').with(environment: ['HOME=/root', 'http_proxy=http://example.com', 'https_proxy=http://example.com']) }
+      end
+
+      describe 'when called with proxy_server and proxy_type parameters' do
+        let(:params) { { proxy_server: 'example.com', proxy_type: 'http' } }
+
+        it { is_expected.to contain_archive('download composer').with(proxy_server: 'example.com') }
+        it { is_expected.to contain_archive('download composer').with(proxy_type: 'http') }
+        it { is_expected.to contain_exec('update composer').with(environment: ['HOME=/root', 'http_proxy=http://example.com', 'https_proxy=http://example.com']) }
+      end
+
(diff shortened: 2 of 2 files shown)


yakatz:
Thank you for your contribution. Can you add tests?

teluq-pbrideau:
@yakatz Sorry for the delay.  I’m not very used to create tests, and I’ve tried, but the class seems unfriendly to easily add tests: There is no default parameters, so to run a test as simple as this:
```ruby
      describe 'works without params' do
        it { is_expected.to compile.with_all_deps }
      end
```
does not works…

Is this PR the right place to refactor the class to set the default parameters (`max_age`,`source`,`path`)?
Or should I create tests with all the params defined in every test cases?
```ruby
      describe 'when called with proxy_server parameters' do
        let(:params) do {
          {
            max_age: '',
            source: '',
            path: '',
            proxy_server: 'http://example.com'
          }
        }
      [... should succeed]
      }
      describe 'when called with proxy_server and proxy_type parameters' do
        let(:params) do {
          {
            max_age: '',
            source: '',
            path: '',
            proxy_type: 'http',
            proxy_server: 'example.com'
          }
        }
      [... should succed]
      }
      describe 'when called with proxy_server without proxy_type parameters' do
        let(:params) do {
          {
            max_age: '',
            source: '',
            path: '',
            proxy_server: 'example.com'
          }
        }
      [... should fail]
      }
```

Or would you suggest another method?

teluq-pbrideau:
Well, I’ve tried even with the definition of every params, and it still fail with :
> error during compilation: Evaluation Error: Error while evaluating a Function Call, Class php::composer::auto_update is private

So I guess I should test the entire class `php::composer` instead of only the changes I made in `php::composer::auto_update` ?

teluq-pbrideau:
@yakatz I’ve tried to make a very basic test for `php::composer`, and also about the changes I made in `php::composer::auto_update`.  Any comment would be appreciated to improve them!

teluq-pbrideau:
I just rebased from master to make sure everything still OK

yakatz (approved):
Looks good to me. Thank you!

PR #743, opened 2025-07-11
add: support for apache-specific extension config on Debian

griggi-ws:
This ensures extension settings are configured in their respective Apache config files as well as the default cli config.

Current behavior:
Extension config file is created in the `apache2` subdirectory by `phpenmod`, but the settings defined in Puppet are not populated.

Tested on Ubuntu 20.04 + ondrej PHP 5.6 (environment that needed fixing), but should function on Debian + other versions, and the paths can be easily overridden.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,7 +1,9 @@
 ---
 fixtures:
   repositories:
-    apt: https://github.com/puppetlabs/puppetlabs-apt.git
+    apt:
+      repo: https://github.com/puppetlabs/puppetlabs-apt.git
+      tag: v10.0.1
     archive: https://github.com/voxpupuli/puppet-archive.git
     inifile: https://github.com/puppetlabs/puppetlabs-inifile.git
     stdlib: https://github.com/puppetlabs/puppetlabs-stdlib.git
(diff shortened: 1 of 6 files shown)


TheMeier:
@griggi-ws thank you for contributing, can you please add some tests?

TheMeier:
@griggi-ws  change `.fixtures.yml` like this https://github.com/voxpupuli/puppet-falco/pull/46/commits/fc3a4bfa1dc2396ec0ea1114993f8be5cf082240

TheMeier (approved):

griggi-ws:
Rebased on latest changes in master to resolve conflicts

.fixtures.yml
```diff
@@ -1,7 +1,9 @@
 ---
 fixtures:
   repositories:
-    apt: https://github.com/puppetlabs/puppetlabs-apt.git
+    apt:
+      repo: https://github.com/puppetlabs/puppetlabs-apt.git
+      tag: v10.0.1
```
kenyon, line 6: Vox Pupuli doesn't support Puppet 7 anymore, so we shouldn't be pinning puppetlabs-apt like this. We dropped Puppet 7 support in #745.

kenyon (changes requested):

Issue #134, opened 2015-11-20
phpversion fact calls "php -v" without checking if php is installed


spotzero:
On the first puppet run, I get the following error:
Could not retrieve fact='phpversion', resolution='<anonymous>': private method `split' called for nil:NilClass

It looks like the reason is that the phpversion fact tries to run "php -v" without checking if php is present.

fpletz:
I don't see this error message. What version of Puppet and Ruby are you using?

fpletz:
I was able to reproduce this and fixes this on master.

PR #768, opened 2026-09-17
Stop the phpversion fact raising without php

tobixen:

**Disclaimer**: The code here is AI-generated, by Claude Opus 5. Having the AI generate pull requests is cheap, reviewing the code is not (though, this one should be really trivial).  I respect that some maintainers may want to close AI-generated pull requests without spending time on explaining why.  The real value I provide here is bug discovery, reproduction and testing.

This is a follow-up on commit f35162553f4dd14902b54d61d85df60cff4490a3 from pull request #767, which reintroduces the problem reported in issue #134.

Claude was prompted to fix the problem with this puppet output:

```
(...)
Notice: /File[/opt/puppetlabs/puppet/cache/lib/facter/phpversion.rb]/content:
--- /opt/puppetlabs/puppet/cache/lib/facter/phpversion.rb    2026-03-20 12:15:28.775782952 +0000
+++ /tmp/puppet-file20260917-169843-1msbgfu    2026-09-17 21:31:11.107207208 +0000
@@ -2,7 +2,7 @@

 Facter.add(:phpversion) do
   setcode do
- output = Facter::Util::Resolution.exec('php -v')
+ output = Facter::Core::Execution.execute('php -v')

  unless output.nil?
(...)
>  Error: Facter: Error while resolving custom fact fact='phpversion', resolution='<anonymous>': Could not execute 'php -v': command not found
```

The trigger is obvious: the php module was being used on nodes that have no php.  I pulled in the master branch in my Puppetfile, the module is pinned back to v12.0.0 there for now and that's sufficient for me - but others will for sure get into trouble later if this change is pushed out in an upcoming release.

Below the line is the AI-generated text saying more or less the same, just with more technical details.

---

f351625 (#767) introduced a regression for every node that does not have php
installed.  `Facter::Util::Resolution.exec` returned `nil` when the command
could not be run; `Facter::Core::Execution.execute` raises
`Facter::Core::Execution::ExecutionFailure` instead.  Facter catches the
exception, so the fact still resolves to `nil`, but it logs an error on every
run of every php-less node:

```
Error: Facter: Error while resolving custom fact fact='phpversion', resolution='<anonymous>': Could not execute 'php -v': command not found
```

That is the same class of noise that the `nil` guard in 6c1e69e was added to
prevent, for #134.

Passing `on_fail: nil` is the direct equivalent of the old `exec` semantics - it
is literally what the deprecated `Facter::Core::Execution.exec` did internally,
and what Facter's own string-form `setcode` does.

The fact resolves to `nil` with and without the fix, since Facter catches the
exception, so the added spec asserts the symptom instead: that nothing is
logged.  It fails on master and passes with the one-line change.

Unrelated to this regression, and left alone here: a php that runs but writes
nothing to stdout still ends in `NoMethodError` on `nil.split`, because the
`unless output.nil?` guard lets `""` through.  Say the word and I will send a
separate patch.

Issue #134 - it was fixed previously, but the latest changes in the master branch reintroduce it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


kenyon (commented):
I think it should be using a `confine` like this: https://github.com/voxpupuli/puppet-letsencrypt/blob/836d38e5526e4b6286d3bfe28f1cd412ee7e2b4e/lib/facter/certbot_version.rb

tobixen:
Fixed

kenyon (approved):

spec/unit/facter/phpversion_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'phpversion fact' do
+  subject(:fact) { Facter.fact(:phpversion).value }
```
bastelfreak: Is this actually executing facter? I don't think this is right. Same for the `load` in line 10. Please compare this to other tests, like https://github.com/voxpupuli/puppet-collectd/blob/master/spec/unit/collectd_real_version_spec.rb

